Issue #1, opened 2020-02-03
Use a Load function for plugin registration


pmazzini:
Plugin loading is a bit cumbersome, it requires a plugin name and factory function for each plugin, e.g.
```
var testSteps = map[string]teststep.TestStepFactory{
	echo.Name:           echo.New,
	slowecho.Name:       slowecho.New,
	example.Name:        example.New,
	cmd.Name:            cmd.New,
	sshcmd.Name:         sshcmd.New,
	randecho.Name:       randecho.New,
	terminalexpect.Name: terminalexpect.New,
}

...
func main() {

...
	// Register TestStep plugins
	for name, tsfactory := range testSteps {
		if err := pluginRegistry.RegisterTestStep(name, tsfactory); err != nil {
			log.Fatal(err)

		}
	}
...
}

```

If we define a loading function that every plugin must implement, we can instead simply do:
```
// teststeps.go
type TestStepLoader func(...) (string, TestStepFactory)

// main.go
var testSteps = []teststep.TestStepLoader{
	echo.Load,
	slowecho.Load,
	example.Load,
	cmd.Load,
	sshcmd.Load,
	randecho.Load,
	terminalexpect.Load,
}
```

The registration functions in the pluginregistry will just need a minimal change, i.e. calling `name, factory := pluginLoader(...)` at the beginning.

PR #2, opened 2020-02-03
[teststep] Use a Load function for plugin registration

pmazzini:
Simplifying plugin registration as suggested in #1.

This is only for the test steps. If it look good, I'll do the same with the other type of plugins.


pmazzini:
It doesn't look like travis failures are related to this PR :s.

insomniacslk:
@pmazzini can you rebase on top of #3 ?

PR #18, opened 2020-02-05
Reporting: initial support for multiple reporters

insomniacslk:
This is the first of a large, API-breaking changes to the reporting contract.
Before this change, we only support a single reporter, that is executed at the
end of each run. This changeset enables specifying multiple reporters (now
called run reporters), and adds the concept of final reporters, i.e. reporters
that are executed when all the runs have completed. This is useful to create
reports based on the results of the whole job instead that on individual runs.

A final reporter for an indefinitely-running job (i.e. with Runs set to
0) will run only when the job is cancelled.
Final reporters are optional, but at least one run reporter has to be
specified.

This commit simply changes the job descriptor to use the new syntax, but
the behaviour is unchanged. Only one run reporter is supported at this
point, and no final reporters yet.

The reporting syntax in the job descriptor should now look like the
followingi (remembering that with this commit only one run reporter and
no final reporters are supported for now):
```
    "Reporting": {
        "RunReporters": [
            {
                "Name": "TargetSuccess",
                "Parameters": {
                    "SuccessExpression": ">80%"
                }
            },
            {
                "Name": "SomeOtherReporter",
                "Parameters": {
                    "SomeParameters": "Blah"
                }
            }
        ],
        "FinalReporters": [
            {
                "Name": "TargetSuccess",
                "Parameters": {
                    "OverallSuccessExpression": ">50%"
                }
            },
            {
                "Name": "AverageTime",
                "Parameters": {
                    "blah": "blah"
                }
            }
        ]
    }
```

### DIFF
--- a/cmds/clients/contestcli-http/start-literal.json
+++ b/cmds/clients/contestcli-http/start-literal.json
@@ -31,8 +31,14 @@
             }
         }
     ],
-    "ReporterName": "TargetSuccess",
-    "ReporterParameters": {
-        "SuccessExpression": ">50%"
-   }
+    "Reporting": {
+        "RunReporters": [
+            {
+                "Name": "TargetSuccess",
+                "Parameters": {
+                    "SuccessExpression": ">80%"
+                }
+            }
+        ]
+    }
 }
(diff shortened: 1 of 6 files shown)


insomniacslk:
fixing tests

PR #28, opened 2020-02-05
Integrate docker run in travis CI

marcoguerri:
Use docker-compose setup to run tests via TravisCI. Export integration tests coverage metrics for pkg, plugins, cmds via codecov.io.

Signed-off-by: Marco Guerri <marcoguerri@fb.com>


.travis.yml
```diff
@@ -1,30 +1,30 @@
+sudo: required
```
insomniacslk: Isn't this ineffective since last year?

marcoguerri: Yes, since they move to 100% virtualized infra. I will remove it.

PR #31, opened 2020-02-06
Reporting called at the end of each run instead of the end of the job

insomniacslk:
This commit fixes an issue where reporting was called at the end of the
job (i.e. after all the desired runs) instead of after every run. The
final reporting will be done through final reporters, which are
currently being implemented.

Before this change, a job with number of runs != 1 would call the
reporter only at the very end (and plugins like TargetSuccess would fail
because it supports only one result, while it will be fed more than
one).
After this change, the reporter will run once per run.

### DIFF
--- a/cmds/clients/contestcli-http/start-literal.json
+++ b/cmds/clients/contestcli-http/start-literal.json
@@ -1,6 +1,6 @@
 {
     "JobName": "test job",
-    "Runs": 1,
+    "Runs": 3,
     "RunInterval": "5s",
     "Tags": ["test", "csv"],
     "TestDescriptors": [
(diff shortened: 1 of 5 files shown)


insomniacslk:
draft pull request, because currently only the last report is returned in the status.

insomniacslk:
CI is passing, this is ready to review (PTAL @marcoguerri and @pmazzini ).

Current limitations: this is a multi-stage change. This commit moves the reporting to the right place, and enables proper multi-run reporting. However the status still only supports one reporter, and the only the last run report will be returned. I will add multiple reports in the status in the next change

PR #36, opened 2020-02-06
make locker configurable

pmazzini:
Moving the lock engine instantiation from the job runner to the main, so that it can be customized.

### DIFF
--- a/cmds/contest/main.go
+++ b/cmds/contest/main.go
@@ -12,6 +12,7 @@ import (
 	"os/signal"
 	"syscall"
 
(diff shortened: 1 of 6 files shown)


insomniacslk:
Looks good to me. Can you  add some CLI output when switching lockers?

pmazzini:
I changed my mind and I am hardcoding it in the main for now, no flags. Same thing that is currently being done with the storage engine. At least anybody can create its own main with a different locking plugin now. I'll add the flags when we need them.

Issue #34, opened 2020-02-06
Persists the targets acquired by the TargetManager


marcoguerri:
Right now there is no source of truth that tells which targets were acquired at the beginning of the job, that can be references for example at events submission time. To have a normalized data model, we need to store targets acquired as well.

xaionaro:
Am I understand correctly that the task is effectively to implement targets-lock-reentrance for a job?

marcoguerri:
This was not meant to be about target locking, but basically storing permanently all the information of targets that we acquire. The target locks backed by MySQL are meant to be ephemeral.

The idea here is instead that we need to have a way to normalize the schema when referencing targets. Right now test events for example are not normalized and it should really be fixed... 

We would have a surrogate key in a targets table (`target_id`, although, this name becomes overloaded and we might think of something else instead) and in test events we'd reference that key.

marcoguerri:
I wouldn't merge the concepts of `targets` and `target_locks` in a single table for example because we need to keep locking independent of the storage backend.

xaionaro:
> I wouldn't merge the concepts of targets and target_locks in a single table for example because we need to keep locking independent of the storage backend.

Sure, I wasn't implying it :)

> Right now test events for example are not normalized and it should really be fixed...

Makes sense. OK.

PR #44, opened 2020-02-14
add locks table

pmazzini:


insomniacslk:
Ah I see this is still a draft, I didn't realize when reading from mobile

pmazzini:
> Should this come together with the SQL locking plugin?

I want to agree on the schema first. I'll implement the plugin afterwards.

marcoguerri:
Apologies for super late review... I have only one concern: what is `target_id` going to be? This is the problem that I was trying to address with #34. Several schema at the moment are not normalized (first and foremost `testevents`). Having a target id to use a foreign key would fix this problem, hence when I mentioned that events should not be the source of truth for which targets we are working on in this job, but instead there should be a dedicate table. How is `target_id` going to be used here?

pmazzini:
I don't think #34 is a hard dependency for this but I may be wrong.

> How is target_id going to be used here?

Each lock request takes a `Target` as argument:
https://github.com/facebookincubator/contest/blob/master/pkg/target/locker.go#L31

And we can use its ID for storing a lock.

As long as the targetmanager fetches the Target with the proper ID, I think it should work. Am I missing something?

pmazzini:
Thoughts?

marcoguerri:
The same target, with the same ID, might be fetched at multiple times, so in practice that cannot be the primary key?

marcoguerri:
Yeah, actually, I think is what we want. We'll just need make sure the framework enforces that IDs of target are unique (I don't think we do that explicitly anywhere, although we treat targets as comparable objects so target equality at the moment means ID, name and FQDN are equal).

Issue #42, opened 2020-02-12
Events API should error out if more that one modifier of the same type is provided


marcoguerri:
The `Fetch(fields []QueryField)` methods accept a list of `QueryFields` (#22 proposes to make it variadic but the same issue would still be relevant), but doesn't complain if two identical modifiers are passed. If that's the case, only the last modifier takes affect and this is very counter intuitive.

marcoguerri:
Fixed by #50 and #22

PR #47, opened 2020-02-17
Forbid query field duplication

xaionaro:
This PR is supposed to be applied over changes of https://github.com/facebookincubator/contest/pull/46.

ITS: https://github.com/facebookincubator/contest/issues/42

### DIFF
--- a/pkg/event/errors.go
+++ b/pkg/event/errors.go
@@ -1,0 +1,31 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+//
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+package event
+
+import (
+	"fmt"
+)
+
+// ErrQueryFieldPassedTwice is returned when QueryFields failed validation due
+// to multiple QueryField which modifies the same field (this is unexpected
+// and forbidden).
+type ErrQueryFieldPassedTwice struct {
+	QueryField QueryField
+}
+
+func (err ErrQueryFieldPassedTwice) Error() string {
+	return fmt.Sprintf("field %T is passed twice", err.QueryField)
+}
+
+// ErrQueryFieldHasZeroValue is returned when a QueryFields failed validation
+// due to a QueryField with a zero value (this is unexpected and forbidden).
+type ErrQueryFieldHasZeroValue struct {
+	QueryField QueryField
+}
+
+func (err ErrQueryFieldHasZeroValue) Error() string {
+	return fmt.Sprintf("field %T has a zero value", err.QueryField)
+}
--- a/pkg/event/frameworkevent/framework.go
+++ b/pkg/event/frameworkevent/framework.go
@@ -7,6 +7,7 @@ package frameworkevent
 
 import (
 	"encoding/json"
+	"fmt"
 	"time"
 
 	"github.com/facebookincubator/contest/pkg/event"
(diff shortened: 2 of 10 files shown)


marcoguerri:
I would have taken a different approach here. Rather than using reflection and pushing to the clients of the query field the responsibility of building  `abstractQueryFields` and the calling `Verify`, I would have modified the query fields to do the following:

* Return an error when applied, if they are applying a default value (this is most likely, if not certainly, not the desired behavior).
* Return an error, when applied, if they are overriding a non-default value, which would surface the cases when multiple `QueryFields` of the same type are specified. 

Thoughts about this approach?

xaionaro:
If I understand correctly and if to be more specific then the approach you propose is to add checks to every QueryField apply-function. So each of those functions will check those zero-value-related conditions. If it is correct then I believe it's a less maintainable solution:
* It requires a lot of code duplication, so any change in logic would require a simultaneous change in a lot of pieces of the code. Which could lead to a human error.
* In my very humble opinion, it is less readable/understandable. If there's a global rule then it should be implemented as a global rule instead of being a local rule copied through each element. If there's a lot of local rules (instead of one global one) then it's harder to understand the initial intention ("should the same rule be applied to a new field?"), harder to read (the code is scattered) and harder to do small fixed without knowing the history of this code. In my opinion everything should be implemented on the level it applies on. Sorry for my English if I explain myself badly here. What I'm trying to say here is formulated by DRY as: "Every piece of knowledge must have a single, unambiguous, authoritative representation within a system".

Besides:
* Checking through zero-value is a kind-of checking through side effects. I believe it is better when something is checked directly. If we need to forbid some field to be mentioned twice then we need to forbid it directly instead of checking the effect on the value. If by some reason a non-zero value will be dropped back to the zero value (in some other logic someday in future) it does not mean we should permit to add a duplicating field after that.
* It's a tiny argument, but still: the proposed by PR approach separates `Verify` from `Apply` which is good (so these functions could be used separately in future).

On the other hand:
* I understand that all these arguments are weak and it's hard for me to give a specific example where the zero-value approach will definitely lead to an error.
* If it is required to require non-zero values (in `QueryField`-s) then the current proposal (by the PR) will also have to utilize `reflect.(*Value).IsZero()`, so the usage of `reflect` will grow.

### A comment about reflection

>  Rather than using reflection and ...

Package `reflect` itself isn't bad. It's bad when a program is written:
* With moving compile-time errors to run-time errors (which often happens when a programmer use `reflect`)
* Unnecessary too abstract (so it is hard to read and maintain).

I don't use any `reflect`-y softness/unstrictness here, so I don't move the scope of errors to run-time. And I don't use anything that is hard to read here, so it's a minimalistic abstraction to comply the DRY principle.

marcoguerri:
I agree that this is a global rule, and distributing the logic across the field is harder to maintain than having the logic implemented in a single place. At the same time, what I think I wanted to express was the following:

* I think less work the done by the clients in this case is desirable. The logic of implementing verification of duplicates is now in the `storage` package. The `storage` package implements an interface which is defined in the `events/...`, and it's a client of the various `event/...` packages. Any other implementer of the same interfaces should be able to re-use the same validation logic, without depending on the storage package.

*  As a client, the `storage` package now has to call `Verify` to perform this check. All other clients will have to do the same. If they forget, this enforcement will not be applied. I think clients should not have to call `Verify`. In my opinion, now that we have validation logic to implement on the query side, clients should decide which modifiers they want, ask for a query with those modifiers, and be given an error if any validation logic which we have decided to implement fails. Agree that if we want to modify the validation logic for the query, we should need to change the logic only in one place.

* This check is meant to protect against cases when one would expect the query to contain a value, and instead it contains another one. This might happen also happens as a result of re-using a query object. So I think that as part of implementing this validation logic, we should prevent also these other cases.

For these reason, I think the API should go into the direction of implementing all validation in the `events/...` packages. Agree to not implement override checks in the `QueryFields`.

xaionaro:
> I think less work the done by the clients in this case is desirable.

Agreed if a "client" is an entity from `./cmds/clients/` in this context.

---

> All other clients will have to do the same. If they forget, this enforcement will not be applied.

If I understand correctly the check exists on the server side only.

Confirmation:

```sh
(
TEMP_DIR="$(mktemp -d)"
trap 'rm -rf "$TEMP_DIR"' EXIT TERM
cd "$TEMP_DIR"
go build github.com/facebookincubator/contest/cmds/clients/contestcli-http
go build github.com/facebookincubator/contest/cmds/contest
objdump -d contestcli-http | grep -c abstractQueryFields
objdump -d contest | grep -c abstractQueryFields
)
```
The output:
```
0
23
```

So this should not make any complications for clients.

Here an example how it works:
```go
	msg := ev.Msg.(api.EventStatusMsg)
	jobID := msg.JobID

	jobReport, err := jm.jobReportManager.Fetch(jobID)
	if err != nil {
		return &api.EventResponse{
			JobID:     jobID,
			Requestor: ev.Msg.Requestor(),
			Err:       fmt.Errorf("could not fetch job report: %v", err),
		}
	}

	// TODO: JobManager relies only on events to rebuild the status of the job.
	// It should fetch a complete list of targets, and render the progress based
	// on events instead.

	// Fetch all the events associated to changes of state of the Job
	jobEvents, err := jm.frameworkEvManager.Fetch(
		frameworkevent.QueryJobID(jobID),
		frameworkevent.QueryEventNames(JobStateEvents),
	)
```
So a client sends to server `ev.Msg` which contains `JobID`. And the server uses this `Fetch` only inside (and without network communication).

However, a question:

* This non-duplication constraint is the property of a query or it is a property of a fetcher?

If I try to answer that, then:
A combination of fields exists only in the `Fetcher`, so we may think it is a property of a fetcher. On the other hand if there will appear another interface with the same signature it will have the same constraint. And it looks like it's a property of a query, so I need to move this `Verify` to `./pkg/event/`. But I still think it should be implemented through `TypeOf`. @marcoguerri : What do you think?

---

> So I think that as part of implementing this validation logic, we should prevent also these other cases.

You mean like if there's a zero-value in a queried field?

marcoguerri:
Sorry, I was not clear in my explanation.  With "client", I meant (probably, not appropriately) an entity which uses a package. 

To me it's a constraint  of a `Query` and other implementations of the same interface should benefit from the same validation logic.

> But still I think it should be implemented through TypeOf instead of duplicating code
Agree.

> calling Verify/Validate from a Fetcher
So, multiple implementations of the `Fetcher` interface will still have to call `Verify` explicitly? If we decide that it's a property of the `Query`, doesn't it make sense to expose an API from the `events` package that performs this validation for all users of the packages, transparently?

xaionaro:
>> calling Verify/Validate from a Fetcher
> So, multiple implementations of the Fetcher interface will still have to call `Verify` explicitly?

Sorry, I already removed that part from my comment. I just realized it is a way to avoid even that.

I will update a PR in few minutes. Let's take a look then :)

xaionaro:
I pushed the change few hours ago. But I also tried to add an unittest which will verify that there's no bad QueryField implementations:

https://github.com/xaionaro-facebook/contest/commit/422a4bdd1b0c51b1e2e8e6e83437e6b7434e5f93

It's very hacky, so I think we should not merge it. But the purpose was to find a way to prevent QueryField implementations which would operate with the same fields as any other QueryField implementations.

marcoguerri:
I would honestly still be in favor of a zero-value based rejection, but feel free to open [xaionaro-facebook@422a4bd](https://github.com/xaionaro-facebook/contest/commit/422a4bdd1b0c51b1e2e8e6e83437e6b7434e5f93) as a separate PR, we'll discuss there.

xaionaro:
> I would honestly still be in favor of a zero-value based rejection

You mean to add a check to every single one apply-function of every QueryField implementation?

marcoguerri:
Probably a variation of that, without having to implement the check itself in each field, but at least delegating to the `QueryField` the responsibility to return the current value in the query, and then enforce that it's zero-value at application time. It has similar problems to what has already been discussed already, so I am not asking that it is implemented this way. Feel free to merge this and re-open `422a4bd` as a separate PR, certainly it will benefit from other's opinions.

xaionaro:
Well, I don't know the context of this project, so if your aesthetic perception says it's a better way I believe it is. Am I understand correctly that the reason you think so is that your approach forbids wrong behavior more explicit and it's harder to make a mistake in the validation?

marcoguerri:
Yes, this is in line with what I had in mind when I mentioned that "the `QueryField` has the responsibility to return the current value in the query" and then something else enforces the check, based on the zero-value. Because after going through more iterations and thinking about this, the zero-value check in my opinion is not a side effect. We want to enforce that:

* A field is not set to zero
* A field is not overwritten

This, regardless of the name of the query field, as you have pointed out above. I like the idea that the `QueryField` still only has one responsibility in #50 . I see the best trade-off in this approach, so if you don't have additional concerns, I would go with #50 .

xaionaro:
> Because after going through more iterations and thinking about this, the zero-value check in my opinion is not a side effect.

Yeah, I was wrong, it seems to be a reliable direct check :)

> I like the idea that the QueryField still only has one responsibility in #50 . I see the best trade-off in this approach, so if you don't have additional concerns, I would go with #50 .

Yeah, I like it more than this initial approach as well. Closing this obsolete PR (#47).

Issue #13, opened 2020-02-04
Make "label" field explicit in test step definitions


insomniacslk:
Currently a test step is implicitly labeled with the plugin name. If the same plugin is used more than once, this leads to a name clash. We agreed on making labels explicit and mandatory

PR #48, opened 2020-02-17
fix!: Make a label to be mandatory for a TestStep

xaionaro:
ITS: https://github.com/facebookincubator/contest/issues/13


pkg/pluginregistry/bundles.go
```diff
@@ -25,7 +25,7 @@ func (r *PluginRegistry) NewTestStepBundle(testStepDescriptor test.TestStepDescr
 	}
 	label := testStepDescriptor.Label
 	if label == "" {
-		label = testStepDescriptor.Name
+		return nil, ErrStepLabelIsMandatory{testStepDescriptor}
```
insomniacslk: Struct initialization without explicit field name should be avoided - I thought we had a linter check in place for that

tests/integ/jobmanager/common.go
```diff
@@ -424,3 +424,23 @@ func (suite *TestJobManagerSuite) TestJobManagerJobCancellationFailure() {
...
 
 }
+
+func (suite *TestJobManagerSuite) TestTestStepNoLabel() {
+	go func() {
+		suite.jm.Start(suite.sigs)
+		close(suite.jobManagerCh)
+	}()
+
+	_, err := suite.startJob(jobDescriptorNoLabel)
```
marcoguerri: Since we are returning `ErrStepLabelIsMandatory`, can we assert error type here?

marcoguerri:
Also, I think we should look into moving some error definitions that we have in `cerrors` into the packages they belong to and actually get rid of that package.

tests/integ/jobmanager/common.go
xaionaro: Yeah, I was going to do so. But it (`errors.As`) was implemented only in Go1.13, while this package is checked by Go1.11 (or so) in CI. So I decided not to add an extra dependency for an assert only.

pkg/pluginregistry/bundles.go
xaionaro: Fixed.

tests/integ/jobmanager/common.go
xaionaro: Fixed.

xaionaro:
> Also, I think we should look into moving some error definitions that we have in cerrors into the packages they belong to and actually get rid of that package (which is the approach used here).

I did not notice this package. Should I move the new error there (while `cerror` still exists)?

marcoguerri:
> I did not notice this package. Should I move the new error there (while cerror still exists)?

I would leave them as is, and instead get rid of `cerrors` (I'll open an issue)

tests/integ/jobmanager/common.go
marcoguerri: Mh.. we should also fix the Travis configuration to use go 1.13, integration test are run in a go 1.13 container. Thank you for spotting this.

xaionaro:
Please note that I checked verb `%v` to `%w` in one place. It will cause less neat error messages in Go<1.13. It will still compile and work, but still.

marcoguerri:
Apologies, I also did not merge this. Could you please look into the merge conflict (trivial json descriptor...). Feel free to merge autonomously afterwards.

xaionaro:
> Apologies, I also did not merge this. Could you please look into the merge conflict (trivial json descriptor...).

No problem, done :)

> Feel free to merge autonomously afterwards.

I don't have privileges in this repository/project.

marcoguerri:
We should fix the privileges problem.

xaionaro:
> We should fix the privileges problem

OK. I've just added myself to the team :)

marcoguerri:
Time to go.

Issue #25, opened 2020-02-05
Move error handling to xerrors


marcoguerri:
In order to support Go 2 proposal for error values (https://golang.org/design/29934-error-values), let's switch to xerrors package.

xaionaro:
The most important stuff is supported out-of-the-box by standard `errors` package in Go>=1.13. We may just use it.

PR #49, opened 2020-02-17
Add initial support for run id in Status

marcoguerri:
This patch adds support for building a status report that takes into
consideration the run id. The events API still does not know anything
about the run id, so the support is not complete, but as soon as it
is added to the query interface, it would be just a matter of adding
a new query parameter. 

Also, the concept of test step index has been removed.

### DIFF
--- a/cmds/contest/test_samples/slowecho.json
+++ b/cmds/contest/test_samples/slowecho.json
@@ -1,0 +1,12 @@
+{
+    "steps": [
+        {
+            "name": "slowecho",
+	        "label": "aslowecho",
+            "parameters": {
+                "sleep": ["3"],
+		"text": ["nel mezzo del cammin"] }
+        }
+    ]
+}
+
--- a/pkg/jobmanager/status.go
+++ b/pkg/jobmanager/status.go
@@ -19,6 +19,14 @@ import (
 	"github.com/facebookincubator/contest/pkg/types"
 )
 
+// JobCompletionEvents gather all event that mark the end of a job
+var JobCompletionEvents = []event.Name{
+	EventJobCompleted,
+	EventJobFailed,
+	EventJobCancelled,
+	EventJobCancellationFailed,
+}
+
 // JobStateEvents gather all event names which track the state of a job
 var JobStateEvents = []event.Name{
 	EventJobStarted,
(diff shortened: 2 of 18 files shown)


cmds/contest/test_samples/slowecho.json
```diff
@@ -0,0 +1,12 @@
+{
+    "steps": [
+        {
+            "name": "slowecho",
+	    "label": "aslowecho",
```
insomniacslk: nit: mixed indentation, can be changed in another PR

cmds/contest/test_samples/slowecho.json
```diff
@@ -0,0 +1,12 @@
...
+            "parameters": {
+                "sleep": ["3"],
+		"text": ["nel mezzo del cammin"] }
```
insomniacslk: nice :)

pkg/jobmanager/status.go
```diff
@@ -38,131 +46,183 @@ var TargetRoutingEvents = []event.Name{
...
 			testevent.QueryEventNames(TargetRoutingEvents),
 		},
 	)
 	if err != nil {
-		return fmt.Errorf("could not fetch events associated to target routing: %v", err)
+		return nil, fmt.Errorf("could not fetch events associated to target routing: %v", err)
 	}
 
-	// Filter only the events emitted by this TestStep, group events by Target
+	targetStatuses := make([]job.TargetStatus, 0)
```
xaionaro: Just out of curiosity: why this way instead of just? --
```
var targetStatuses []job.TargetStatus
```

Is there any reason to get a non-nil value in case of an empty slice?

pkg/jobmanager/status.go
```diff
@@ -38,131 +46,183 @@ var TargetRoutingEvents = []event.Name{
...
+	// TODO: This query will have to include the runID, which can be found in the coordinates
+	testEvents, err := jm.testEvManager.Fetch(
+		[]testevent.QueryField{
+			testevent.QueryJobID(coordinates.JobID),
+			testevent.QueryTestName(coordinates.TestName),
+			testevent.QueryTestStepLabel(coordinates.TestStepLabel),
+		},
+	)
+	if err != nil {
+		return nil, fmt.Errorf("could not fetch events associated to test step %s: %v", coordinates.TestStepLabel, err)
```
xaionaro: It would be better to use `%w` instead of `%v`.

pkg/jobmanager/status.go
```diff
@@ -38,131 +46,183 @@ var TargetRoutingEvents = []event.Name{
...
-		// Update the TargetStatus object associated to the Target. If there is no
-		// TargetStatus associated yet, append it
-		var currentTargetStatus *job.TargetStatus
-		for index, candidateTargetStatus := range currentTestStepStatus.TargetStatus {
-			if *candidateTargetStatus.Target == *testEvent.Data.Target {
-				currentTargetStatus = &currentTestStepStatus.TargetStatus[index]
+		// Update the TargetStatus object associated to the Target.
+		// If there is no TargetStatus associated yet, append it
+		var targetStatus *job.TargetStatus
+		for index, candidateStatus := range targetStatuses {
```
xaionaro: Very subjective, just ignore it :) :

I think it would've been better to sort `routingEvents` by `Target` instead of doing a linear search on each step. Also I would move this whole `for` to a separate function (like `(job.TargetStatuses).ToTargetStatuses()`).

pkg/jobmanager/status.go
```diff
@@ -38,131 +46,183 @@ var TargetRoutingEvents = []event.Name{
...
+	}
+
+	// Build a map of events that should not be rendered as part of the status (e.g. routing events)
 	skipEvents := make(map[event.Name]bool)
 	for _, eventName := range TargetRoutingEvents {
 		skipEvents[eventName] = true
 	}
+	filteredTestEvent := []testevent.Event{}
+	for _, event := range testEvents {
+		if _, skip := skipEvents[event.Data.EventName]; !skip {
```
xaionaro: Just a note:
If you use it this way, then you can define the map as `map[event.Name]struct{}` instead of `map[event.Name]bool`. It will consume less memory.

pkg/runner/job_runner.go
```diff
@@ -256,10 +271,54 @@ func (jr *JobRunner) Run(j *job.Job) ([][]*job.Report, []*job.Report, error) {
 	return allRunsReports, finalReports, nil
 }
 
+// GetCurrentRun returns the run which is currently being executed
+func (jr *JobRunner) GetCurrentRun(jobID types.JobID) (*types.RunID, error) {
```
xaionaro: Just out of curiosity:

Why `*types.RunID`? There could never be a case so the value will be `nil` while error is not `nil`. So the error could be used as the detector of validity.

pkg/runner/job_runner.go
```diff
@@ -256,10 +271,54 @@ func (jr *JobRunner) Run(j *job.Job) ([][]*job.Report, []*job.Report, error) {
...
+
+func (jr *JobRunner) emitEvent(jobID types.JobID, eventName event.Name, payload interface{}) error {
+	payloadJSON, err := json.Marshal(payload)
+	if err != nil {
+		jobLog.Warningf("could not encode payload for event %s: %v", eventName, err)
+		return err
+	}
+
+	rawPayload := json.RawMessage(payloadJSON)
+	ev := frameworkevent.Event{JobID: jobID, EventName: eventName, Payload: &rawPayload, EmitTime: time.Now()}
```
xaionaro: Just a note:

May be we should consider to use `time.Now().UTC()` instead of `time.Now()` through the whole ConTest. It will prevent possible problems with timezones.

pkg/runner/job_runner.go
```diff
@@ -256,10 +271,54 @@ func (jr *JobRunner) Run(j *job.Job) ([][]*job.Report, []*job.Report, error) {
...
+	if err := jr.eventManager.Emit(ev); err != nil {
+		jobLog.Warningf("could not emit event %s: %v", eventName, err)
+		return err
+	}
+	if err := jr.eventManager.Emit(ev); err != nil {
```
xaionaro: **Important**

Why do you `Emit` it twice?

pkg/jobmanager/status.go
```diff
@@ -38,131 +46,183 @@ var TargetRoutingEvents = []event.Name{
...
+func (jm *JobManager) BuildStatus(currentJob *job.Job) ([]job.RunStatus, error) {
+
+	runStatuses := make([]job.RunStatus, currentJob.Runs)
+	for runID := uint(0); runID < currentJob.Runs; runID++ {
+		runCoordinates := RunCoordinates{JobID: currentJob.ID, RunID: types.RunID(runID)}
+		runStatus, err := jm.BuildRunStatus(runCoordinates, currentJob)
+		if err != nil {
+			return nil, fmt.Errorf("could not rebuild run status for run %d: %v", runID, err)
+		}
+		runStatuses[runID] = *runStatus
```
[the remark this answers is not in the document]
marcoguerri: Makes sense, will amend.

pkg/jobmanager/status.go
```diff
@@ -190,41 +246,65 @@ func (jm *JobManager) status(ev *api.Event) *api.EventResponse {
...
 	jobStatus := job.Status{
-		Name:       currentJob.Name,
-		StartTime:  startTime,
-		EndTime:    reportTime,
-		JobReport:  jobReport,
-		TestStatus: make([]job.TestStatus, 0, len(currentJob.Tests)),
+		Name:      currentJob.Name,
+		StartTime: startTime,
+		EndTime:   endTime,
+		State:     string(jobEvents[len(jobEvents)-1].EventName),
```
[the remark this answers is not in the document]
marcoguerri: Yeah, will fix, thanks.

pkg/runner/job_runner.go
marcoguerri: Bug.

pkg/runner/job_runner.go
marcoguerri: The object returned by `Time.Now()` has timezone information embedded (and then we store in UTC)? And as far as I know there is no concept of naive time object in Go (contrary to Python), so time comparison should always be done between objects with timezone information.

pkg/jobmanager/status.go
insomniacslk: I agree - this could be incorporated in #25

pkg/runner/job_runner.go
insomniacslk: the only risk could be related to storing/retrieving the time string to/from database, but we store the non-naive timestamp so this should never become an issue. I agree that UTC is better though, it could be done in the future if we identify more important gaps related to this

xaionaro: > And as far as I know there is no concept of naive time object in Go (contrary to Python), so time comparison should always be done between objects with timezone information.

"Possible problems with timezones" that I referred to -- are problems while converting time to different data-formats (go:Time -> JSON -> DB -> go:Time and so on).

Never mind. It was just a (may be stupid) thought. I just used to think that timezones is a sources of vague bugs.

marcoguerri: We store in `TIMESTAMP` MySQL format which interprets incoming time object as belonging to  the timezone which is set for the current connection (which is the client timezone). It's then persisted as UTC and returned again in the timezone set within the connection for the query. I'd need to look into how the connector works if the incoming time object is encoded in a timezone which does not match the connection set in the timezone (I assume it's handled transparently, but I'd need to confirm).

As for encoding/decoding, we would still serialize with timezone information (e.g. RFC3339) and de-serialize accordingly. I won't address this now, but let's revisit if we really think this approach is fragile.

insomniacslk: Time zones are a source of HUGE problems, not just vague, and you're right to worry about them :) What we mean here is that the timestamps are already stored in a non-naive way, which eliminates (minimizes?) the risk of timezone-related problems. Additionally, Go makes it really hard to work with naive timestamps, which helps keep our sanity

marcoguerri:
Mistakenly pushed my `.golangci.yml` :( Will fix.

insomniacslk:
after that, feel free to merge

Issue #24, opened 2020-02-05
Improve error reporting when TestStep tries to emit an event which is not allowed


marcoguerri:
When a TestStep tries to emit and event which is not allowed, the emitter fails with a generic error. It should be more specific and give pointers to be able to identify which TestStep actually failed. I have a refactor of the events emission control path ready, so this issue should be addressed there.

xaionaro:
Khm. How to reproduce this? I've just made a quick look and I see no event name checks at all at the moment.

Event is just added (and then flushed):
```go
	r.frameworkEventsLock.Lock()
	r.buffFrameworkEvents = append(r.buffFrameworkEvents, event)
	if len(r.buffFrameworkEvents) >= r.frameworkEventsFlushSize {
		doFlush = true
	}
	r.frameworkEventsLock.Unlock()
```

Issue #14, opened 2020-02-04
Plugin events registration to become case insensitive


insomniacslk:
To reproduce: declare a plugin name with any capitalized letter, and use it lower-case in the test description. The job will fail with "TestStep does not have any event associated"

Issue #10, opened 2020-02-04
Disallow plugins named "contest"


insomniacslk:
We are going to use plugin names for event namespaces, and "contest" is going to be a reserved name for the framework itself

xaionaro:
If I understand correctly the task then it could be implemented as an additional check here:

https://github.com/facebookincubator/contest/pull/60/files#diff-84f3bd06e0d39e53607b8726fc4ec61fR15

xaionaro:
The draft is in PR#64

PR #64, opened 2020-03-05
Refactor plugin factories

xaionaro:
This PR is implemented on top of changes of: https://github.com/facebookincubator/contest/pull/60

Refactored the way plugins are registered. It allowed to deduplicate code and add common rules/constraints/code for any type of plugin.

The initial idea was introduced in commit c919e3fa08779427830c48de5adb97c7de8eca75 and this commit just completes it.

Closes:
 * https://github.com/facebookincubator/contest/issues/14
 * https://github.com/facebookincubator/contest/issues/10
 * https://github.com/facebookincubator/contest/issues/24

BREAKING CHANGE: changed the convention of plugin factories, all plugins are need to be updated to comply *Factory interfaces.

---

A brief summary:

An abstract factory is introduced here. An abstract factory could be registered in the registry, be used from the registry, validated and more. Highlights of the changes:
* Added interfaces `abstract.Factory` and `abstract.Factories` (and `abstract.FactoryWithEvents` for validation purposes).
* For each specific-factory-slice type added method `ToAbstract()` (which wraps/boxes each factory to `abstract.Factory` to get the factories represented as `abstract.Factories`)
* Replaced all separate `map`-s of `PluginRegistry` to an unified ~~map~~ ~~`map[reflect.Type]map[string]abstract.Factory`~~ ~~`map[FactoryType]map[string]abstract.Factory`~~ slice `[]map[string]abstract.Factory`.
* Updated `(*PluginRegistry).New*` methods to get a factory through unified logic. But the their API is mostly preserved (the only change is simplified the usage of `NewTestStep`: it returns all the required information and there's no need to call `NewTestStepEvents` anymore).
* Added logic for #14 , #10 and #24 to the unified method `RegisterFactory`.
* Updated all the factories. So now instead of separate `New()`, `Name` and `Events()` it's all collected as methods of empty struct `Factory`. So this collection of properties could be addressed through a factory. And each of this `Factory`-ies implements `abstract.Factory` (which contains the common logic for any factory).
* Added some unit-tests.

A factory registration flow:
* `(*PluginRegistry).RegisterFactory(&SpecificFactory{})` registers the factory

A factory usage flow:
* `(*PluginRegistry).NewSpecificEntity(imlementationName)` -->
* `(*PluginRegistry).Factory(specificEntityFactoryType, imlementationName)()`


marcoguerri:
Could you elaborate why we need an additional layer of abstraction  to solve #14 #10 #24? Can we not just reduce duplication in `PluginRegistry`, within the `PluginRegistry` itself to check for plugin name, force lowercasing, etc.? This to me seems a significant amount of additional complexity for the problem we are trying to solve.

xaionaro:
> Can we not just reduce duplication in PluginRegistry, within the PluginRegistry itself to check for plugin name, force lowercasing, etc.?

Previously `PluginRegistry` for every type of factory had a separate `Register*` method. So It was required to add the logic to multiple methods. Now it's deduplicated.

As side effect it allows to do some tiny things like this (see `possible values: %s`):
```
		log.Fatalf("Implementation '%s' is not found (possible values: %s)",
			factoryImplName, factories)
```
I just type `factories` here and receive all possible values. Because the code is deduplicated as it's very cheap to add such functionality, so I just add it :)

Also for example I added `main_test.go` which checks all the factories. And it wasn't required to check different types one by one.

> This to me seems a significant amount of additional complexity for the problem we are trying to solve.

1. I saw repeatings related to factories in multiple places, so it seemed to be a strong pattern within ConTest, so I just coded it as a pattern instead of separate implementations. Yes, it's a new abstraction. But what is the criteria when to start an abstraction? I usually use the rule of three as one of the criterias: if the same implementation repeats three (or more) times, then it should be deduplicated. While here things were repeated by 5 times. And when I was implementing a MySQL-based locker I had a choice to add the 6th time or to deduplicate it.
2. The most of added lines are comments, visual changes (like rolling-out a two-lines into 5 lines for more readability), added error-types (to do not use indistinct error types) and changes in the broken (before this changes IIRC) test.
3. In my opinion, the complexity is added only into plugin registration mechanism (to be support an abstract factory).

> Also, please try to give a brief summary of the PR on top. This, together with #60, is a significant amount of code, and it would be appreciated if you could help give meaningful review.

I will try to add more description in 24 hours.

xaionaro:
> This, together with #60, is a significant amount of code

I suggest to look at a single commit only (instead of PR diff): https://github.com/facebookincubator/contest/commit/c919e3fa08779427830c48de5adb97c7de8eca75

xaionaro:
> Also, please try to give a brief summary of the PR on top.

Updated.

xaionaro:
> too many problems that derive from lighthearthed use of reflection

For example? (specifically to this case where I don't use any `FieldByName` and other soft logic stuff)

insomniacslk:
> @insomniacslk :
> 
> > too many problems that derive from lighthearthed use of reflection
> 
> For example? (specifically to this case where I don't use any `FieldByName` and similar stuff). It's just only `(*reflect.Value).Implements()` (which is only for additional checks which in turn are not required, actually) and `reflect.TypeOf()` (just to operate with `reflect.Type` instead adding a new type with almost the same meaning). I can totally eliminate `reflect` from here and preserve the logic (and don't use `unsafe` or any other cheating of course), just the way with reflect it's more customary (and therefore readable) in my humble opinion.

Reflection tends to make code harder to read, and to reason with, because it makes code dynamic. Some compile-time constraints and checks are lost, and in general this makes it a less-than-ideal candidate unless there's really no other option. I personally strongly believe that some duplication is better than invisible "magic" (which is what reflection adds here, even if you're not making a wild use of it), because even if some code appears to be dumbly replicated, it's always obvious what it does, to the human and to the compiler.


> > I'm not going to merge something like this in the ConTest codebase
> 
> Will you reconsider if I will just remove `reflect` at all from this PR?

If you mean reconsider unifying the factories without dynamic or unsafe code, probably yes (I don't know what you have in mind there). I am not against removing code duplication of course, I'd just like to see it done in a readable and runtime-safe manner.


> > ConTest is meant to be simple, understandable
> 
> I sincerely was moving into direction of easy-to-understand. Because too flat code makes personally me waste a lot of my brain resources to process (it's like to read an assembly code instead of structured C-code).

I believe I understand what's your rationale, and I would have done the same a few years ago. I think I've just been bitten too much by more readable but more obscure code, where the intention seems obvious but it's not necessarily so (for example, I consider the `reportIfError` example in #60 less obvious than its Go-idiomatic counterpart with `if err := ... err != nil`)

xaionaro:
> Reflection tends to make code harder to read, and to reason with, because it makes code dynamic.

I'm aware.

> Some compile-time constraints and checks are lost, and in general this makes it a less-than-ideal candidate unless there's really no other option.

Actually the added dynamic is limited by simple type-switch. As I said `Implements` was added just as additional check (just-in-case, just because I don't like softness/dynamicness of reflect) that most likely will never be required/used. So if you don't like this `Implements` we can just safely remove it (and declare a couple of extra methods to be unexported). As you can see in the used code, all the dynamicness is filtered with a type-switch in the code, so there's no way for a reflect-caused mistake.

That's why I asked: "`For example? (specifically to this case where I don't use any FieldByName and similar stuff)`". I used reflect quite carefully (but yeah: it still could've been done much more careful, I just didn't want to spend too much time on this).

> I personally strongly believe that some duplication is better than invisible "magic"

Agree.

> (which is what reflection adds

In general, agree.

> here

Specifically in this case I don't see how it adds softness/unstrictness/dynamicness or obscurity.

> because even if some code appears to be dumbly replicated, it's always obvious what it does, to the human and to the compiler.

No, because:
* Maintaining duplicated code causes copy&paste errors. Sometimes a modification is not copied into a copy (while there's 5-7 copies).
* It just requires more brain resources to parse and understand all the code. It's just too many lines of code to make it obvious. As with assembly code. The minimal subset of assembly language is much simpler/KISSer than C, but read an assembly code is harder (because the code is represented on a lower lever than the task is).

> If you mean reconsider unifying the factories without dynamic or unsafe code

No, I don't mean "reconsider". I mean just remove `reflect` without adding other red flags. Just to prove that `reflect` was/is not adding any reflect problems here. I will just demonstrate that the code with reflect is equivalent to a safe code without reflect.

> I think I've just been bitten too much by more readable but more obscure code,

Well, this thing has two ends. There's other end which we can see in scientific programs (that are ran on HPC clusters). A lot of physicist took `no abstractions if possible` as the rule, and it's not maintainable and not readable as well. So obviously it's always should be a wise decision instead of maximalic rules like `avoid abstractions at all cost` or `do everything as abstraction`. And in my opinion the rule of three (deduplicate the code if it repeats at least three times) is close to a wise approach in the most cases (but still not always).

> (for example, I consider the reportIfError example in #60 less obvious than its Go-idiomatic counterpart with if err := ... err != nil)

This is a bad example, because this is the case where I have large doubts about the way I used myself. But still I have to stand my ground to have a dialog in which I can give a clearer answer which way is better. But, actually, I won't argue specifically about `reportIfError` :)

---

Anyway. I will just remove "reflect" from here in few hours :)

xaionaro:
Removed `reflect`.

marcoguerri:
I'll try to explain what I meant with my comment:

>> Can we not just reduce duplication in PluginRegistry, within the PluginRegistry itself to check for plugin name, force lowercasing, etc.

What I have in mind does not use abstract factories, but I think is addresses the logic duplication problem, where dangerous and relevant. In fact, this is what we have been discussing if I did not misunderstand: more verbose code is fine, logic duplication is not (then we might still debate what is acceptable verbosity and what is not).

To me, working with factories that are associated to concrete types is easier to reason about, and the cost is acceptable.

So, I have kept in my solution `TargetManagerLoader`, `TestStepLoader` and kept `RegisterTargetManager`, `RegisterTestStep`, etc. Basically no change from the plugin point of view.

The registration logic however is now centralized:
```
func (r *PluginRegistry) RegisterTestStep(l test.TestStepNewLoader) error {
	return r.registerPlugin(l())
}
```

where register plugin is the following:
```
// RegisterPlugin registers a generic plugin
func (r *PluginRegistry) registerPlugin(name string, f interface{}, p interface{}) error {
	if factory, ok := f.(test.TestStepFactory); ok {
		tsp, ok := p.(test.LoaderParams)
		if !ok {
			return fmt.Errorf("test step loader not associated to test step params")
		}
		log.Infof("Loading a new test step plugin: %s, %v", name, tsp.Events)
		r.Plugins[name] = factory
	} else if _, ok := f.(target.TargetManagerFactory); ok {
		log.Infof("Loading a new target manager plugin: %s", name)
	}
        [...]
}
```

If the argument is that `interface{}` is less safe than `Factory interface` as it's the weakest type you could conceive, probably yes, but it's only used internally to the plugin registry, everything externally works with strong types. I see also above you commented:

>> I just guess interface{} may make similar feelings to reflect-induced ones for an external developer.

If confined internally to the `PluginRegister`, I would not share that opinon.

The validation logic can be centralized in `registerPlugin`. `Register*` are minimal trampolines for a centralized logic, and yes, work with concrete types, so are duplicated. The registration logic in main that iterates over all the different types of plugins, registers them, checks the return value, can be centralized into the plugin registry itself, so that other `main` have minimal room for error (e.g. checking correctly the returned error).

To address your points:
>> Replaced all separate map-s of PluginRegistry to an unified map map[reflect.Type]map[string]abstract.Factory map[FactoryType]map[string]abstract.Factory.
Can be done here as well, everything that ends up in this map is known to be safe and recognized at compile time thanks to the external interface of `PluginRegistry`.

>> Updated (*PluginRegistry).New* methods to get a factory through unified logic. But the their API is mostly preserved (the only change is simplified the usage of NewTestStep: it returns all the required information and there's no need to call NewTestStepEvents anymore).

Also can happen with the proposal above.


>> Added logic for #14 , #10 and #24 to the unified method RegisterFactory.
Same.


>> Updated all the factories. So now instead of separate New(), Name and Events() it's all collected as methods of empty struct Factory. So this collection of properties could be addressed through a factory. And each of this Factory-ies implements abstract.Factory (which contains the common logic for any factory).

I think here the same as above applies. If there are parameters that are specific to one type of plugin, that will be conveyed in the registration function, returned by `Load`. The internal `registerPlugin` will consider those parameters only if relevant (see `TestStep` example above).

To conclude, I would find the solution proposed above safer at compile time, easier to reason about, and capable of protecting against the same logic duplication traps that the proposal with `Factory interface` is trying to solve.

xaionaro:
@marcoguerri :

Thank you for the proposal :)

The answer is below:

---

# Primary thoughts

> ```go
> 	return r.registerPlugin(l())
> ```
> ```go
> func (r *PluginRegistry) registerPlugin(name string, f interface{}, p interface{}) error {
> 	if factory, ok := f.(test.TestStepFactory); ok {
> 		tsp, ok := p.(test.LoaderParams)
> ```

What difference I see here (in comparison to the PR's approach):
1. You have no unite type of a map for factories and that would mean you will have more duplication. Let's call this point "**Separate maps**", I will comment it below.
2. Use 3 separate types (string, unknown, unknown) to define one factory instead of 1 type (which just collects this properties together). It does not affect logic in any way, it's only about coding style. I will comment it below (let's call this point "**Coding style**")

### Separate maps

I tried this way and:
* The resulting code is less readable for me (but I guess more readable for you and Andrea) and showed more cognitive complexity by `gocognit`. However I guess myself and `gocognit` are just a corner case here, so we may just ignore that.
* This implementation also does allow to deduplicate other stuff, like `String()` for factories slice (to handle `flag`-s itself, see `main.go`), or like initialization of a factory depending on a flag (parse a flag call the right factory, see `parseFactoryInfo`), or factories initialization process (fill the plugin registry, see `setupPluginRegistry`), or like call-the-factory logic (like `(*PluginRegistry).NewTestFetcher`), or like checking plugin name duplication (since every map is separate), or like setting values to maps. May be something else (I did not check)... Or may be I miss something and you have an idea how to deduplicate these stuff using no common type.

### Coding style

If we forget about separate maps (which I tried and wasn't satisfied). The proposed approach is effectively the same as the PR's approach. You may use the same type `Factory` as in my approach and redeclare `registerPlugin` this way:
```go
func (r *PluginRegistry) registerPlugin(factory interface{ Name() string }) error {
```
So the logic remains absolutely the same, it just a way how to put it as a Go code. And the way with an interface is more Go-way (we may see this way as a strong pattern in the standard packages), IMHO. And if we are not a lazy coder an unnamed interface and will create an interface (which IMHO should be called `AbstractFactory` or `abstract.Factory`).

### In total

* If you believe we should separate maps and face the duplicating consequences — may be so. If my argument below did not convince you, then OK, let's try to separate back those maps (I don't like it, but I believe the consensus is one your side and it's a fair/good/effective way to solve this).
* Also if you believe the non-interfacive coding style in this task is better and my arguments haven't convinced you, then OK, once more quorum is on your side. But if you are still in doubt then may be this post will help: https://github.com/facebookincubator/contest/pull/60#issuecomment-597789597

---

# Secondary thoughts

> If confined internally to the `PluginRegister`, I would not share that opinion.

I agree. And the same argument was (and is) valid for dynamicness/softness (which is stricter) was proposed in the PR (including the version with `reflect.Type`).

> The registration logic in main that iterates over all the different types of plugins, registers them, checks the return value, can be centralized into the plugin registry itself

I did not understand this. Could you clarify please?

While you use specific separate types you cannot unify this logic. Or you just starts to use `interface{}` as the common type. But I don't see why unnamed (not self-explaining) unrestricted (fully dynamic) `interface{}` is better than `abstract.Factory`. Just because you don't need to declare a type? (so to reduce the verbosity of the code?)

>> Replaced all separate map-s of PluginRegistry to an unified map map[FactoryType]map[string]abstract.Factory.
> 
> Can be done here as well, everything that ends up in this map is known to be safe and recognized at compile time thanks to the external interface of `PluginRegistry`.

I did not understand this as well :(

You use separate `map` here. If you unify them you will just convert this proposal to the PR's proposal (with difference in coding style only, see above).

> To conclude, I would find the solution proposed above safer at compile time

If without making maps separate then why? Or just because of separate maps?

marcoguerri:
Concerning a single map vs separate map, I agree with the problems you mention. We want to reduce code duplication, and having separate maps doesn't seem to be the way to go. And I agree that `interface{}` is a downgraded version of abstract.Factory.

However, if I look at the factory interface
```
type Factory interface {
   UniqueImplementationName() string
}
```
and I ask myself: "What does this type tell me about a factory?". The answer is "It has a name", which is my opinion not much. So, the way I see it is that `Factory` is not really a very much upgraded version of `interface{}` either. And if ["Interfaces in Go provide a way to specify the behavior of an object: if something can do this, then it can be used here."](https://golang.org/doc/effective_go.html#interfaces), well, yes, but not because it's a factory, but because it represents something which has a name. So in my mind that interface is not a `Factory`, it's a `Name` interface. Then, it will have to be type-asserted to a plugin specific factory anyway later. My concern is that if `Name()` becomes the base interface for a factory to reduce code duplication, and we use this approach elsewhere, then we'll have an explosion of very thin interfaces that do not carry really much information. They are a slightly upgraded version of `interface{}`.

If the factory was:
```
type Factory interface {
   Name() string()
   New() plugins.Plugin
}
```
where `Plugin` was the interface shared by all plugins, then to me it would be much easier to understand this code. But I am having a hard time accepting the `abstract.Factory` interface, as to me that seems to be a stretch, introduced to remove code duplication, but which makes the type hierarchy and the code more difficult to understand.  At this point, I would be inclined to use `interface{}` within the internal boundaries of `PluginRegistry`, in a very confined space.

> The registration logic in main that iterates over all the different types of plugins, registers them, checks the return value, can be centralized into the plugin registry itself

>> I don't understand this. Could you clarify please? 

Here I was not clear, but I simply meant that there could be a `RegisterPlugins([]TargetManagerFactory, []TestStepFactory, ...)`, that would remove from the main the need for repeated `if err != nil`, which can be a source of bugs. It's just shuffling code around, doesn't solve any code duplication issue.

>> While you use specific separate types you cannot unify this logic (until contracts will be ready). Or you just propose to start to use interface{} as the common type? If so then I don't see why an unnamed (not self-explaining) unrestricted (fully dynamic) interface{} is better than abstract.Factory. Just because you don't need to declare a type? (so to reduce the verbosity of the code?)

To sum up:
* `abstract.Factory` in my opinion does not define the interface for a Factory. It's a `Name` interface, it doesn't build anything. If some other object happens to have a `Name`, then we have two choices:
   * We reuse the same `abstract.Factory` interface. And at that point, what is it going to become? Will it still be an `abstract.Factory`? No, probably it will become a `Name` interface, and at it's going to be shared among concrete implementations that have nothing to do with each other (apart from having a name), so it wouldn't be a safe candidate for registering a plugin.
   * We add another interface, which exposes `Name() string` again. And we do it all over again, until we have an explosion of duplicated interfaces that share the same signatures, but have a different name.

Either of the above to me seem bad alternatives. Worse than having some minimal code duplication and use of  `interface{}` confined internally to `PluginRegistry`.

> Replaced all separate map-s of PluginRegistry to an unified map map[FactoryType]map[string]abstract.Factory.

>> Can be done here as well, everything that ends up in this map is known to be safe and recognized at compile time thanks to the external interface of PluginRegistry.

>>> I don't understand this as well :(

>>> You use separate map here. If you unify them you will just convert this proposal to the PR's proposal (with difference in coding style only, see above).

Here I wasn't clear enough, but I was assuming to have a `map[string]interface{}`.
`main` would register a `TargetMangerFactory`, a `TestFetcherFactory` as it does now, not an `abstract.Factory` (or the way I think about it, an `abstract.Name`), which is not clear what type it is.

> To conclude, I would find the solution proposed above safer at compile time
>> If without making maps separate then why? Or just because of separate maps?
Because we don't register an `abstract.Factory` (or, as I keep on thinking about it, an `abstract.Name`), which apart from having a name, we don't really know what it is. We register a `TargetManagerFactory`, `TestFetcherFactory`.

xaionaro:
> and I ask myself: "What does this type tell me about a factory?". The answer is "It has a name", which is my opinion not much.

Agreed. I had the same feeling while writing it.

> So, the way I see it is that `Factory` is not really a very much upgraded version of interface{} either. 

Well. At least only real factories implements this interface, so it's very easy to find possible implementations, so the code is more readable in my subjective perception. Also access to the name is still important and it is used multiple ways in the PR. And one more: it's just objective thinking (which is natural for a human): "we need some factory (interface with good self-explaining name), here it is your factory (specific implementation which implements the interface)". I mean in our heads, a name **is** a property of a factory (and not a separate independent entity), so it seems to be more natural to make it so in the code as well.

> My concern is that if `Name()` becomes the base interface for a factory to reduce code duplication, and we use this approach elsewhere, then we'll have an explosion of very thin interfaces that do not carry really much information.

Agreed. I don't like it as well.

> If the factory was:
> ```
> type Factory interface {
>    Name() string()
>    New() plugins.Plugin
> }
> ```
> where `Plugin` was the interface shared by all plugins, then to me it would be much easier to understand this code.

I really like this interface. But I'm not sure if I'll be able to introduce it into ConTest. But I'll try :)

Right now is see next problems:
* But still `plugins.Plugin` will be very abstract, may be it will be even an empty interface. But I think it's OK. Anyway in this way we provide more information than it is proposed in my PR and it is better.
* `New()` requires different sets of arguments. May be I'll be have to write it as `New(...interface{}) plugins.Plugin`. But it makes the process of processing this arguments a little-bit more difficult on the factories' side. I will try different approaches to solve this problem (may be a neat way will be found).

> But I am having a hard time accepting the `abstract.Factory` interface, as to me that seems to be a stretch, introduced to remove code duplication, but which makes the type hierarchy and the code more difficult to understand. At this point, I would be inclined to use `interface{}` within the internal boundaries of `PluginRegistry`, in a very confined space.

I understand the point with type hierarchy (was just "require and use implementations", become "require interfaces and use implementation of the interface"). But I don't understand why `interface{}` is better than `abstract.Factory`. If we move `abstract.Factory` to be confined within `pluginregistry` would you agree that it's better than just `interface{}`?

> It's just shuffling code around, doesn't solve any code duplication issue.

It reduces amount of code you need to parse while working with the code. It's basically a syntax sugar in specifically this case (in the plugin registration function of `main.go`). I agree it's a very minor thing and in Go culture is considered to be better to just repeat the same `err != nil` a lot of times. Let's assume there's no usefulness of this "deduplication" in specifically this case, there's still other places where it was more useful :)

> It's a `Name` interface, it doesn't build anything. If some other object happens to have a `Name`, then we have two choices:
> We reuse the same `abstract.Factory` interface. ...
> And we do it all over again, until we have an explosion of duplicated interfaces that share the same signatures, but have a different name.

One of the reasons to name it `UniqueImplementationName` instead of just `Name` is to prevent occasional collisions (so it's not just a name, but an implementation name). Other reason is to emphasize the uniqueness (it's not just a random name, but a name with a specific purpose -- to be used as the identifier). So it won't collide with other interfaces/implementations. However I agree it's not the best reasons to name methods like that. But still this way we don't have the problems you mention here.

> Here I wasn't clear enough, but I was assuming to have a `map[string]interface{}`.
main would register a `TargetMangerFactory`, a `TestFetcherFactory` as it does now, not an `abstract.Factory` (or the way I think about it, an `abstract.Name`), which is not clear what type it is.

If I understood correctly then you cannot do that. Otherwise factories of different types but with the same name will collide with each other.

---

# In total

I agree with the most concerns you mention above (but not with the trivial-interfaces-explosion).

I agree I should dig more into direction:
> ```
> type Factory interface {
>    Name() string()
>    New() plugins.Plugin
> }
> ```

Let me try. May be we will find our consensus there :)

Also an important question: do you have any concerns about `abstract.Factory` if it will be confined within `pluginregistry`?

marcoguerri:
To clarify, I also think that 

```
type Factory interface {
   Name() string()
   New() plugins.Plugin
}
```
Is going to be very abstract and is going to lead to the same discussion as here. My point was: if that abstraction fits well into the plugin architecture, it would be the right thing to do. If it doesn't fit, forcing it into the framework in my opinion is not the right thing to do. Note that this approach will also relax the constraints imposed to the plugin developers, requiring that the factory returns a generic `plugins.Plugin`, which, as we both agree, is going to be very abstract.

xaionaro:
> Is going to be very abstract and is going to lead to the same discussion as here.

Am I understand that you still believe the way with separate maps is the better one?

insomniacslk:
I am catching up with this now, apologies for the delay

xaionaro:
> I am catching up with this now, apologies for the delay

Don't waste time. I'm about to close this PR (if the answer to my last question to Marco is "yes"). It does not bring any essential functional to ConTest and produces too much discussion.

insomniacslk:
> > I am catching up with this now, apologies for the delay
> 
> Don't waste time.

It's still worth taking a look and understanding why this is generating a lot of discussion

insomniacslk:
I will respond only to a few points because this conversation is massive and not worth nit-picking every single thing.

I agree that `interface{}` is not better than `abstract.Factory`, but I also agree that `abstract.Factory` just provides a name (with a convoluted function name), is not really a factory, and doesn't help on clarity and behaviour enforcement. Given that the pros (having a named type) are in my opinion outweighed by the cons (a deeper, non-enforcing abstraction layer; and the risk of a false sense of security given by a named type), I'd recommend going with anonymous interfaces confined in a single function instead.

I'm not convinced by the `gocognit` argument. We are in the field of opinions here, so take it for what it is. `gocognit` (and you) complain that that code is harder to read because it has repetitions. However my point is that that code is simple and very understandable, yet duplicated. If I see it once, I'll spend 10 seconds reading and understanding it. If I see it twice, I'll spend one second and say "ah yeah I know what this is", so the cognitive complexity reported by the tool should not be followed religiously in my opinion.

I also think that `Factory` with `Name` and `New` doesn't solve the problem, and we should not spend more time investigating into it.

I also believe that the code with separate maps is more readable, and safer to reason with for the average Go programmer. I also acknowledge the pros of having a single map, but I believe they are still outweighed by their cons.

Overall, we need to find a balance across our opinions, and an acceptable agreement. I would stop with nit-picking and line-by-line arguments and replies. We need to do what's best for our users, which is really giving them a usable, easy to understand, safe framework. Let's accept to do some step back, realize that both proposals have pros and cons, pick the one that creates less disagreement, and then disagree and commit.

xaionaro:
>  If I see it once, I'll spend 10 seconds reading and understanding it. If I see it twice, I'll spend one second and say "ah yeah I know what this is", so the cognitive complexity reported by the tool should not be followed religiously in my opinion.

Personally I have the opposite effect. My brain wants to do as you say, but I have to enforce it to re-read and to re-interpret the code, because there could be hidden differences and the reason of a bug (I want to find) could be there. I believe it's subjective.

> I'd recommend going with anonymous interfaces confined in a single function instead.

1. You meant confined in single package? Because with unified maps it will be used in multiple functions.
2. As I said I use the method to extract the name multiple times. With empty interfaces I won't be able to do that. But anonymous non-empty interface is usually considered as less maintainable (especially if it is used multiple times). What do you think if I will declare a private named interface?

> I also acknowledge the pros of having a single map, but I believe they are still outweighed by their cons.

OK, I think tries to reach a consensus here does not worth of time will be wasted. Closing the PR.

Issue #75, opened 2020-03-10
Persist test step descriptors


insomniacslk:
We currently don't store the fetched test step descriptors, which makes it hard to debug, audit, and resume jobs.

marcoguerri:
@insomniacslk , this can be closed, right?

insomniacslk:
Yes, thanks for noticing

PR #76, opened 2020-03-10
Persist test descriptors to the storage layer

insomniacslk:
Currently we don't store the fetched test step descriptors, which makes
it hard to debug, audit, and resume jobs. This is the first building
block to achieve the above.

See #75 .


insomniacslk:
This is ready for review. Unfortunately the diff is quite large because I had to change a few interfaces, that are touching many files. However the actual modifications are quite scoped, and the size of the diff is mainly due to adjusting the points where those functions are called.

Additionally, I had to separate job descriptor and test step descriptors in the API, because `job.Request` does not include the representation of the test step descriptors *after* they are fetched, and for how the framework works, they would not be known at that stage.

insomniacslk:
fixing the merge conflicts after #73

marcoguerri:
Apologies, did not think that #73 would cause conflicts with this, it was already accepted for one week and I wanted to get rid of it. I'll get to review this asap today.

plugins/storage/rdbms/report.go
```diff
@@ -163,3 +165,25 @@ func (r *RDBMS) GetJobReport(jobID types.JobID) (*job.JobReport, error) {
 		FinalReports: finalReports,
 	}, nil
 }
+
+// StoreTestSteps persists a list of test step descriptors to the database.
+func (r *RDBMS) StoreTestSteps(jobID types.JobID, steps []*test.TestStepDescriptor) error {
```
marcoguerri: `RDBMS` now supports transactions, and will need to lock all the `Exec`, `Query` and row scanning operations, in case there is an active transaction, so there should be:

```
r.lockTx()
defer r.unlockTx()
```

at the beginning of `StoreTestSteps`. If it wasn't for having to protect rows scanning, this could be hidden behind an interface which wraps `Query` and `Exec`, but unfortunately that's not enough, everything needs to be locked and serialized.

The same applies to `GetTestSteps` below.

insomniacslk:
no worries, merge conflicts are normal :)

plugins/storage/rdbms/report.go
marcoguerri: After going through the whole patch, I am not sure I understand why we need `StoreTestSteps`? When we invoke `StoreJobRequest` we already have the `[][]*test.TestStepDescriptor` (for all tests), so we should be able to persist everything we need in the `job` table via `StoreJobRequest`?

insomniacslk: I've changed approach a couple of times and `StoreJobRequest` is probably a leftover. Let me fix the merge conflicts first, then I'll go through the diff again to remove what should not be there. Sorry!

pkg/job/request.go
```diff
@@ -23,13 +24,13 @@ type Request struct {
 // RequestEmitter is an interface implemented by creator objects that
 // create Request objects
 type RequestEmitter interface {
-	Emit(request *Request) (types.JobID, error)
+	Emit(jobRequest *Request, testDescriptors [][]*test.TestStepDescriptor) (types.JobID, error)
```
xaionaro: **Tidiness (feel free to ignore)**:

Thought 1: I see that those step descriptors are strongly connected to `job.Request` and `job.Job`. Basically they behave in this process as a part of `job.Request` and `job.Job`.
Thought 2: Even in the table schema, `teststeps` are declared as a field of a `job`.
Thought 3: And, if I understand correctly, the initial task is just to make `job.Job` serializable (so it could be stored somewhere). While current approach (if I understand it correctly) is to convert it to two different entities `job.Request` and `job.TestStepDescriptors` and then just pass this couple through the whole chain to save it (and then to extract it the same way back).

So the point is: may be it would be a better way to just make `job.Job` serializable instead (to preserve enough information within to make it serializable and unserializable). I mean a design property of a `job.Job` (serializability) should be implemented as a code of `job.Job` instead of adding extra entities. But it's just a wild idea which wasn't thought-out 🙂

**Just a note (feel free to ignore):**

I was trying to understand what is `job.Request` better. It uses some thing called `EventRequestor`. Which is copied from `api.Event`. Which is passed as the argument to `(*JobManager).start()`, which is called from `(*JobManager).handleEvent()`. Which is called from `(*JobManager).Start()` using message from channel `(*API).Events`, which received messages from `(*API).SendEvent()`. Which is called from `(*API).SendReceiveEvent()`. Which is called from 4 placed, but let's follow the branch which leads us to `(*API).Start()`. It constructs the `api.Event` (mentioned earlier) using `EventRequestor` passed as the argument (to `(*API).Start)`. `(*API).Start()` is called from two places, let's go to the one in httplistener.go. It constructs the `EventRequester` this way: 
`requestor := api.EventRequestor(r.PostFormValue("requestor"))`. Which is initially generated by client from flag `r`. My point is:
* It would be nice to leave a comment what is this `Requestor` in the lines where `job.Request` is constructed 🙂, someday...
* Also it's just not too easy to load all this chain into the head. It may appear to be a good idea to re-think mechanism of passing request-details at all. For example **if** we don't want to redesign it, then may be (it's an extremely humble idea) we should consider request contexts (like [this](https://godoc.org/github.com/labstack/echo#Context), but much simpler) instead of passing such variables manually, which is subjectively much easier to follow (because it's more usual way for a web-service in Go in my subjective perception, so **you always know what it is**) and also a much more flexible way (provides a lot of other possibilities for future needs, like request-logger, request-metrics, request-profiling and other stuff when we will need it). It's just another wild thought (feel free to ignore it as well).

---

Even if you find a sane seed here, let's don't waste time on discussing it right now (just will think it through when there will be no time pressure).

insomniacslk: Thanks for your review :) answers inline


> So the point is: may be it would be a better way to just make `job.Job` serializable instead (to preserve enough information within to make it serializable and unserializable). I mean it a property of a `job.Job` to be serializable by design, so it should be implemented as a code of `job.Job` instead of adding extra entities. But it's just a wild idea which wasn't thought-out, yet 

Yes, this is related to my comment above, https://github.com/facebookincubator/contest/pull/76#issuecomment-600256929 . This is something I would consider for the future, but for the purposes of this PR (persisting full state), the change would be significant, and we'd rather have this sooner than later. I have cut #78 for the purpose

> * It would be nice to leave a comment what is this `Requestor` in the lines where `job.Request` is constructed slightly_smiling_face, someday...

the job requestor is a top-level concept in ConTest, I guess we should update the README to make it more explicit rather than commenting it in every point where this may look ambiguous

> * Also it's just not too easy to load all this chain into the head. It may appear to be a good idea to re-think mechanism of passing request-details at all.

It is not indeed, I had some issue keeping track of this myself. This code evolved over time and would definitely benefit from simplifying this part. I have created #79 for that

insomniacslk: if that sounds OK, I'll go ahead and merge this

insomniacslk:
merging this

pkg/job/request.go
xaionaro: > Yes, this is related to my comment above

Whoops, sorry, I've missed it.

> the job requestor is a top-level concept in ConTest, I guess we should update the README to make it more explicit rather than commenting it in every point where this may look ambiguous

It's just an opinion (feel free to ignore), but I believe a coder won't look into README every time when he sees something unclear. It would be nice to have the code readable without reading README.

insomniacslk: > It's just an opinion (feel free to ignore), but I believe a coder won't look into README every time when he sees something unclear in the code. 

I agree with what you say in general, and I'm for self documenting code, but I think it's different here. What I meant here is that the requestor is a top level concept used in many parts of the code, and I don't think it's worth documenting it where it's called, but rather where it's defined, and in the documentation (just like target managers, test fetchers and the like). Let me know if I'm misunderstanding what you mean here

PR #77, opened 2020-03-18
Honor go.mod

marcoguerri:
I realize only now that by default go modules is ignored:

```
go get: warning: modules disabled by GO111MODULE=auto in GOPATH/src;
	ignoring go.mod;
	see 'go help modules'
```

staticcheck v2020.1.3 introduces one additional failure on possible
nil pointer dereferece which I would like to fix separately wrt #73
(putting a guard for nil value before the reference does not seem
to make the linter happy).

In any case, we shoud honor go.mod and update depdencies in a controlled
manner.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ global:
 env:
   global:
     - COMPOSE_VERSION=1.25.4
+    - GO111MODULE=on
   jobs:
     - TEST_SUITE=tests
     - TEST_SUITE=linters
--- a/.travis/linters.sh
+++ b/.travis/linters.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -exu
 
-go get -u github.com/golangci/golangci-lint/cmd/golangci-lint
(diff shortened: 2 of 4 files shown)


marcoguerri:
Well, linters do not run in docker, so not enough.

insomniacslk:
Looks good, just a linter failure on the `cd` that was added:
```
/home/travis/.travis/functions: line 115: cd: facebookincubator/contest: No such file or directory
```

marcoguerri:
Yeah, I haven't gotten it right yet.

marcoguerri:
Ok, `golangci-lint` simply breaks when using gomod. I'll make sure at least it's enabled in Docker, we'll keep on using latest `golangci-lint` for the moment, and I actually fixed the false positive in #73.

marcoguerri:
Travis sets `GO111MODULE=auto` by default, which are not retained by Docker while building the layers of the container. So, gomod when running integration or e2e works fine. Closing this for now.

Issue #78, opened 2020-03-20
Assess unifying Job information and making it serializable


insomniacslk:
Follow-up to #76 . Job descriptor and test step descriptors are separate entities, and the design would be nicer if they were both encapsulated in a single job object instead. To keep in mind that test step descriptors are not known at job object creation, because they are fetched at a later stage, and they would have to be added to the job object at a later stage (and before inserting them in the DB)

PR #82, opened 2020-03-24
Unify job and test descriptors in job structures

insomniacslk:
In this commit I've unified job and test descriptors in the `job.Request` object, as also hinted by @xaionaro in #76.

### DIFF
--- a/pkg/job/job.go
+++ b/pkg/job/job.go
@@ -54,6 +54,10 @@ type Job struct {
 	// unlimited, are specified.
 	RunInterval time.Duration
 
+	// TestDescriptors is the string form of the fetched test step
+	// descriptors.
+	TestDescriptors string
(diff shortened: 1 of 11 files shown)


insomniacslk:
This is still work in progress, please do not review yet

insomniacslk:
I am making other changes in parallel, and I should probably split the two commits of this PR in two separate commits - the first one can already go, the second requires another intermediate change. I will force-push to remove the second commit

insomniacslk:
This is ready to review now, please take a look

marcoguerri:
Well, no, this unifies everything in `JobRequest`, probably the intention of #78 is to make `Job` itself serializable. Nevermind.

insomniacslk:
That's as you say, #78 is slightly different and more laborious

PR #83, opened 2020-03-30
Merge job events interfaces

insomniacslk:
This commit simplifies the amount of interfaces we use for emitting and
retrieving job-related events. This is in preparation of adding more
methods to access more job-related data.

### DIFF
--- a/pkg/storage/jobevents.go
+++ b/pkg/storage/jobevents.go
@@ -1,0 +1,21 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+//
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+package storage
+
+import (
+	"github.com/facebookincubator/contest/pkg/job"
+)
+
+// JobEventEmitterFetcher wraps all the job-related emitters and fetchers.
+type JobEventEmitterFetcher struct {
+	JobRequestEmitterFetcher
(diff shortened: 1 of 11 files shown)


pkg/storage/jobevents.go
```diff
@@ -0,0 +1,24 @@
...
+// JobEventEmitterFetcher wraps all the job-related emitters and fetchers.
+type JobEventEmitterFetcher struct {
+	JobRequestEmitterFetcher
+	JobReportEmitterFetcher
+}
+
+// NewJobEventEmitterFetcher creates a new emitter/fetcher object for job events
+func NewJobEventEmitterFetcher() job.EventEmitterFetcher {
+	return JobEventEmitterFetcher{
+		JobRequestEmitterFetcher{},
```
xaionaro: It's totally not important at all. Just I don't understand why it was required to explicitly write `JobRequestEmitterFetcher{},` and `JobReportEmitterFetcher{},`. Since it's not a pointer or/and interface it will be initialized with a zero value anyway.

insomniacslk: you are right, I was just reusing the existing style in the file, but it makes no sense and I'll change that. The whole function is not necessary, to be honest, but for this change I'm trying to touch as little as I can

insomniacslk: I'm removing it from other places

PR #85, opened 2020-04-02
Change type of test step Param to wrap json.RawMessage

tfg13:
The goal is to allow JSON structures as parameters, which are further
unmarshalled by the specific plugin using them. This can be useful for a multitude of reasons.
In my case, the plugin needs to make calls to another service, and this service as a ton of parameters, and new ones get added all the time. It would be very cumbersome to manually add support for all of them in my plugin. If we allow structures, we can validate the input at job ingestion time in the plugin, but don't have to hardcode any specific params.

### DIFF
--- a/pkg/test/parameter.go
+++ b/pkg/test/parameter.go
@@ -15,15 +15,16 @@ import (
 	"github.com/facebookincubator/contest/pkg/target"
 )
 
-// NewParam inititalizes a new Param object from a string
-func NewParam(s string) Param {
+// NewParam inititalizes a new Param object directly from a string.
+// Note that no validation is performed if the input is actually valid JSON.
+func NewParam(s string) *Param {
 	var p Param
 	p.RawMessage = json.RawMessage(s)
-	return p
+	return &p
 }
 
 // Param represents a test step parameter. It is initialized from JSON,
-// and can be either a strings or more complex JSON structures.
+// and can be a string or a more complex JSON structure.
 // Plugins are expected to know which one they expect and use the
 // provided convenience functions to obtain either the string or
 // json.RawMessage representation.
@@ -36,6 +37,9 @@ func (p Param) IsEmpty() bool {
 	return p.String() == ""
 }
 
+// String returns the parameter as a string. This helper never fails.
+// If the underlying JSON cannot be unmarshalled into a simple string,
+// this function returns a string representation of the JSON structure.
 func (p Param) String() string {
 	var str string
 	if json.Unmarshal(p.RawMessage, &str) == nil {
@@ -44,9 +48,10 @@ func (p Param) String() string {
 	// can't unmarshal to string, return raw json
 	return string(p.RawMessage)
 }
-
-func (p *Param) UnmarshalJSON(data []byte) error {
-	return json.Unmarshal(data, &p.RawMessage)
+// JSON returns the parameter as json.RawMessage for further
+// unmarshalling by the test plugin.
+func (p Param) JSON() json.RawMessage {
+	return p.RawMessage
 }
 
 // Expand evaluates the raw expression and applies the necessary manipulation,
--- a/pkg/test/parameter.go
+++ b/pkg/test/parameter.go
@@ -15,6 +15,13 @@ import (
 	"github.com/facebookincubator/contest/pkg/target"
 )
 
+// NewParam inititalizes a new Param object from a string
+func NewParam(s string) Param {
+	var p Param
+	p.RawMessage = json.RawMessage(s)
+	return p
+}
+
 // Param represents a test step parameter. It is initialized from JSON,
 // and can be either a strings or more complex JSON structures.
 // Plugins are expected to know which one they expect and use the
--- a/pkg/test/parameter.go
+++ b/pkg/test/parameter.go
@@ -20,31 +20,26 @@ import (
 // Plugins are expected to know which one they expect and use the
 // provided convenience functions to obtain either the string or
 // json.RawMessage representation.
-// Simple strings can contain simple function expressions that can
-// be expanded, for example to include the hostname of the test target.
-type Param json.RawMessage
+type Param struct {
+	json.RawMessage
+}
 
 // IsEmpty returns true if the original raw string is empty, false otherwise.
 func (p Param) IsEmpty() bool {
 	return p.String() == ""
 }
 
-// String returns the parameter as a string. This helper never fails.
-// If the underlying JSON cannot be unmarshalled into a simple string,
-// this function returns a string representation of the JSON structure.
 func (p Param) String() string {
-	var s string
-	if err := json.Unmarshal(p, &s); err == nil {
-		return s
+	var str string
+	if json.Unmarshal(p.RawMessage, &str) == nil {
+		return str
 	}
-	// full JSON
-	return string(p)
+	// can't unmarshal to string, return raw json
+	return string(p.RawMessage)
 }
 
-// JSON returns the parameter as json.RawMessage for further
-// unmarshalling by the plugin.
-func (p Param) JSON() json.RawMessage {
-	return json.RawMessage(p)
+func (p *Param) UnmarshalJSON(data []byte) error {
+	return json.Unmarshal(data, &p.RawMessage)
 }
 
 // Expand evaluates the raw expression and applies the necessary manipulation,
@@ -54,7 +49,7 @@ func (p *Param) Expand(target *target.Target) (string, error) {
 		return "", errors.New("parameter cannot be nil")
 	}
 	// use Go text/template from here
-	tmpl, err := template.New("").Funcs(getFuncMap()).Parse(string(*p))
+	tmpl, err := template.New("").Funcs(getFuncMap()).Parse(p.String())
 	if err != nil {
 		return "", fmt.Errorf("failed to parse template: %v", err)
 	}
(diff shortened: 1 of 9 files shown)


tfg13:
The current version has issues.

The main one is that the compiler currently allows doing things like `len(params.GetOne("foo")), which will no longer do what it looks like, as GetOne returns JSON.

Also we call unmarshal on every single access.

tfg13:
This needs a custom `UnmarshalJSON` as well. It looks like the normal unmarshalling thinks Params is `[]byte` (which is kinda correct) and expects base64 data only:
`could not validate TestFetcher fetch parameters: illegal base64 data at input byte 0`

pkg/test/step.go
```diff
@@ -36,7 +36,7 @@ func (t TestStepParameters) Get(k string) []Param {
 func (t TestStepParameters) GetOne(k string) Param {
 	v, ok := t[k]
 	if !ok || len(v) == 0 {
-		return Param("")
+		return Param{}
```
insomniacslk, line 39: We should probably return a pointer now

pkg/test/parameter_test.go
```diff
@@ -23,7 +24,8 @@ func TestParameterExpand(t *testing.T) {
 		[4]string{"name={{ .Name }}, id={{ .ID }}", "blah", "12345", "name=blah, id=12345"},
 	}
 	for _, x := range validExprs {
-		p := Param(x[0])
+		var p Param
+		p.RawMessage = json.RawMessage(x[0])
```
tfg13, line 28: This is kinda ugly. Not sure there is a better way

pkg/test/step.go
tfg13: No strong opinion, I can change it

pkg/test/parameter_test.go
tfg13: We could offer `NewParam(string)` to make this nicer. Yay/Nay?

pkg/test/step.go
insomniacslk: Sorry, I meant that the API should not return a copy but a value, since it's now a struct. It probably won't make a big difference from a performance perspective, but it's definitely not a pure literal anymore

pkg/test/parameter.go
```diff
@@ -7,23 +7,46 @@ package test
...
-// Param represents a test step parameter. It is initialized from a string,
-// which can be either a host name, a function expression, or just any regular
-// string. It enables hostname and function expression validation and a few
-// other validation operations. This is meant to simplify the way plugin access
-// and use test step parameters.
-type Param string
+// NewParam inititalizes a new Param object from a string
+func NewParam(s string) Param {
+	var p Param
+	p.RawMessage = json.RawMessage(s)
```
tfg13, line 21: This takes random string content and pretends it is valid JSON.
This only works because `Param.String()` below returns the raw message as string if `Unmarshal` fails.

Would it be a better to:
- Check if the input is valid json, then accept
- Otherwise tread as random string and marshal into json. This can fail. Not sure what to do then.

pkg/test/parameter_test.go
tfg13: I did this now

pkg/test/step.go
tfg13: ok doing it

insomniacslk:
apologies, I will not have time today to review this, will try tomorrow

pkg/test/parameter.go
marcoguerri: Shouldn't `Param` expose both `AsString` (with `string, error` signature) which fails if we cannot unmarshal into string, and `AsJSON` (with `json.RawMessage, error` signature) which instead fails if we can marshal into string? 

 The plugin should have clear expectations on what the parameters should look like. It might event decide to accept any of the two formats, then try both `AsString` and `AsJSON`, and be given a clear signal by `Param` based on the return values on which one is valid, and act accordingly.

tfg13: We can do that, but it makes calling AsString() very annoying for the simple usecase, so it clutters the plugin code. I think @insomniacslk wanted to prevent that. The current API with `GetOne` goes into a similar direction: It never fails, but might return empty strings.
I am not married to this, I can totally change it. What does @insomniacslk think?

insomniacslk: I personally think `AsString` should just return `string(p.RawMessage)`, without trying to interpret it (at the end of the day, the JSON subdocument is a string, and it's again up to the plugin to interpret it if it's not a plain string).

What would `AsJSON` take as input, and return? I'm not sure I understand what its role is
CC @marcoguerri

> I think @insomniacslk wanted to prevent that.

In general, yes, if I can choose I prefer to move the complexity away from users and plugin developers, but of at the end of the day we have to do what makes sense, if there's no better way it's OK to revert the above principle

pkg/test/parameter.go
```diff
@@ -7,23 +7,46 @@ package test
...
+	var str string
+	if json.Unmarshal(p.RawMessage, &str) == nil {
+		return str
+	}
+	// can't unmarshal to string, return raw json
+	return string(p.RawMessage)
+}
+
+func (p *Param) UnmarshalJSON(data []byte) error {
+	return json.Unmarshal(data, &p.RawMessage)
```
insomniacslk, line 49: won't this cause a stack overflow with UnmarshalJSON calling json.Unmarshal, calling again UnmarshalJSON etc?

tfg13: No it will call `UnmarshalJSON` of the type you are trying to unmarshal to, so on `json.RawMessage`.

But this comment was still very valuable, it made me to a bit more extensive testing, and I realized we don't need this after all, the default logic will end up calling the unmarshal function of `json.RawMessage` anyway. I think this was required only in a previous version. It's gone now :)

pkg/test/parameter.go
tfg13: Decided to keep it simple for now, as this should only be relevant for unit tests

pkg/test/parameter.go
```diff
@@ -7,23 +7,52 @@ package test
...
 
-// Param represents a test step parameter. It is initialized from a string,
-// which can be either a host name, a function expression, or just any regular
-// string. It enables hostname and function expression validation and a few
-// other validation operations. This is meant to simplify the way plugin access
-// and use test step parameters.
-type Param string
+// NewParam inititalizes a new Param object directly from a string.
+// Note that no validation is performed if the input is actually valid JSON.
+// This helper is primarily intended for tests.
```
insomniacslk, line 20: since it's only for tests, this should be private

pkg/test/parameter.go
```diff
@@ -7,23 +7,52 @@ package test
...
+	if json.Unmarshal(p.RawMessage, &str) == nil {
+		return str
+	}
+	// can't unmarshal to string, return raw json
+	return string(p.RawMessage)
+}
+// JSON returns the parameter as json.RawMessage for further
+// unmarshalling by the test plugin.
+func (p Param) JSON() json.RawMessage {
+	return p.RawMessage
```
insomniacslk, line 55: should `RawMessage` also be private? With the helpers it seems we don't need direct access to it

pkg/test/parameter.go
insomniacslk: nevermind, I see that this is used by  `sshcmd`. Maybe we should not hint that this is mainly for tests?

tfg13: Hmm yeah I did not think about that when I wrote the commit, I might just remove that remark.

Alternatively, I can change `sshcmd`, it does not have to create a `Param` there, it might just a string. But I guess it is more consistent if we keep it `Param`.

pkg/test/parameter.go
tfg13: Oh, can you control the visibility of this kind of struct embedding somehow?

tfg13: I am reading the interwebs and it seems I can't make the embedded field private :/

pkg/test/parameter.go
tfg13: I removed the comment

pkg/test/parameter.go
insomniacslk: not the embedded field itself, but if we have a `JSON` method this could become a non-anonymous field. However this is good as it is, I was just wondering if we could handle that differently, I'll merge now

marcoguerri:
Looks like it will support what we want to do, however I am still wondering: when do we expect ` json.Unmarshal(p.RawMessage, &str) ` in the code below:
```
func (p Param) String() string {
	var str string
	if json.Unmarshal(p.RawMessage, &str) == nil {
		return str
	}
	// can't unmarshal to string, return raw json
	return string(p.RawMessage)
}
```

to succeed?

tfg13:
I expect it to succeed for simple string test params that were properly imported from JSON:
```json
"foobar": "hello"
```
`GetOne("foobar")` will now return a Param that has a quoted string inside (`"hello"`), which will unmarshal to the unquoted string `hello`, which is then returned here.

The main cases where unmarshalling fails are
- The parameter is not a simple string (happens for substructs)
- The param is not valid JSON (happens mostly in unit tests which can do `NewParam("hello")`, but technically should do `NewParam("\"hello\"")`)

marcoguerri:
All right, I missed that unmarshalling into `json.RawMessage` a json doc consisting of a quoted string would leave it quoted. The rest makes sense, thanks.

PR #88, opened 2020-04-08
Add -wait flag to wait for completion after start command

ahmomo:
This change adds -wait flag to be used with "start" command. After the job is created, the CLI will keep polling for the result, and only after the job is completed it will print the result and exit.

### DIFF
--- a/cmds/clients/contestcli-http/main.go
+++ b/cmds/clients/contestcli-http/main.go
@@ -15,8 +15,12 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"strconv"
+	"time"
 
 	"github.com/facebookincubator/contest/plugins/listeners/httplistener"
+	"github.com/facebookincubator/contest/pkg/api"
+	"github.com/facebookincubator/contest/pkg/jobmanager"
 )
 
 // Unauthenticated, unencrypted sample HTTP client for ConTest.
@@ -31,11 +35,13 @@ import (
 
 const (
 	defaultRequestor = "contestcli-http"
+	jobWaitPoll      = 10 * time.Second
 )
 
 var (
 	flagAddr      = flag.String("addr", "http://localhost:8080", "ConTest server [scheme://]host:port[/basepath] to connect to")
 	flagRequestor = flag.String("r", defaultRequestor, "Identifier of the requestor of the API call")
+	flagWait      = flag.Bool("wait", false, "After starting a job, wait for it to finish, and exit 0 only if it is successful")
 )
 
 func main() {
@@ -45,6 +51,8 @@ func main() {
 		fmt.Fprintf(flag.CommandLine.Output(), "command: start, stop, status, retry, version\n")
 		fmt.Fprintf(flag.CommandLine.Output(), "  start\n")
 		fmt.Fprintf(flag.CommandLine.Output(), "        start a new job using the job description passed via stdin\n")
+		fmt.Fprintf(flag.CommandLine.Output(), "        when used with -wait flag, stdout will have two JSON outputs\n")
+		fmt.Fprintf(flag.CommandLine.Output(), "        for job start and completion status separated with newline\n")
 		fmt.Fprintf(flag.CommandLine.Output(), "  stop int\n")
 		fmt.Fprintf(flag.CommandLine.Output(), "        stop a job by job ID\n")
 		fmt.Fprintf(flag.CommandLine.Output(), "  status int\n")
@@ -77,26 +85,57 @@ func run(verb string) error {
 			return fmt.Errorf("failed to parse job descriptor: %v", err)
 		}
 		params.Add("jobDesc", string(jobDesc))
+		resp, err := request(verb, params)
+		if err != nil {
+			return err
+		}
+		fmt.Println(resp)
+
+		if *flagWait {
+			fmt.Fprintf(os.Stderr, "\nWaiting for job to complete...\n")
+			parsedData := &api.ResponseDataStart{}
+			parsedResp := &httplistener.HTTPAPIResponse{Data: parsedData}
+			if err := json.Unmarshal([]byte(resp), parsedResp); err != nil {
+				return fmt.Errorf("cannot decode json response: %v", err)
+			}
+			jobID := parsedData.JobID
+			params.Set("jobID", strconv.Itoa(int(jobID)))
+			
+			resp, err = wait(params, jobWaitPoll)
+			if err != nil {
+				return err
+			}
+			fmt.Println(resp)
+		}
 	case "stop", "status", "retry":
 		jobID := flag.Arg(1)
 		if jobID == "" {
 			return errors.New("missing job ID")
 		}
 		params.Set("jobID", jobID)
+		resp, err := request(verb, params)
+		if err != nil {
+			return err
+		}
+		fmt.Println(resp)
 	case "version":
 		// no params for protocol version
 	default:
 		return fmt.Errorf("invalid verb: '%s'", verb)
 	}
+	return nil
+}
+
+func request(verb string, params url.Values) (string, error) {
 	u, err := url.Parse(*flagAddr)
 	if err != nil {
-		return fmt.Errorf("failed to parse server address '%s': %v", *flagAddr, err)
+		return "", fmt.Errorf("failed to parse server address '%s': %v", *flagAddr, err)
 	}
 	if u.Scheme == "" {
-		return errors.New("server URL scheme not specified")
+		return "", errors.New("server URL scheme not specified")
 	}
 	if u.Scheme != "http" && u.Scheme != "https" {
-		return fmt.Errorf("unsupported URL scheme '%s', please specify either http or https", u.Scheme)
+		return "", fmt.Errorf("unsupported URL scheme '%s', please specify either http or https", u.Scheme)
 	}
 	u.Path += "/" + verb
 	fmt.Fprintf(os.Stderr, "Requesting URL %s with requestor ID '%s'\n", u.String(), *flagRequestor)
@@ -107,12 +146,12 @@ func run(verb string) error {
 	fmt.Fprintf(os.Stderr, "\n")
 	resp, err := http.PostForm(u.String(), params)
 	if err != nil {
-		return fmt.Errorf("HTTP POST failed: %v", err)
+		return "", fmt.Errorf("HTTP POST failed: %v", err)
 	}
 	defer resp.Body.Close()
 	body, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
-		return fmt.Errorf("Cannot read HTTP response: %v", err)
+		return "", fmt.Errorf("Cannot read HTTP response: %v", err)
 	}
 	fmt.Fprintf(os.Stderr, "The server responded with status %s", resp.Status)
 	var indentedJSON []byte
@@ -120,12 +159,12 @@ func run(verb string) error {
 		// the Data field of apiResp will result in a map[string]interface{}
 		var apiResp httplistener.HTTPAPIResponse
 		if err := json.Unmarshal(body, &apiResp); err != nil {
-			return fmt.Errorf("response is not a valid HTTP API response object: '%s': %v", body, err)
+			return "", fmt.Errorf("response is not a valid HTTP API response object: '%s': %v", body, err)
 		}
 		// re-encode and indent, for pretty-printing
 
 		if err != nil {
-			return fmt.Errorf("cannot marshal HTTPAPIResponse: %v", err)
+			return "", fmt.Errorf("cannot marshal HTTPAPIResponse: %v", err)
 		}
 
 		buffer := &bytes.Buffer{}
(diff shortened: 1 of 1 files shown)


cmds/clients/contestcli-http/main.go
```diff
@@ -149,12 +183,40 @@ func run(verb string) error {
...
 		indentedJSON = buffer.Bytes()
 	}
 	// this is the only thing we want on stdout - the JSON-formatted response,
 	// so it can be piped to other tools if desired.
-	fmt.Println(string(indentedJSON))
-	return nil
+	return string(indentedJSON), nil
 }
+ 
+func wait(params url.Values) (string, error) {
```
insomniacslk, line 195: could you make the poll interval a parameter for this function? It will make it easier to test

cmds/clients/contestcli-http/main.go
```diff
@@ -149,12 +183,40 @@ func run(verb string) error {
...
+		if fmt.Sprintf("%v", jobState) == fmt.Sprintf("%v", eventName) {
+			finished = true
+			break
+		}
+	}
+
+	if finished {
+		return resp, nil
+	}
+	return wait(params)
```
insomniacslk, line 221: to my knowledge, the Go compiler optimizes tail calls only in a few case, and there is no plan to expand this further. So this call will grow the stack, and while this may be premature optimization, I think we may still prefer iteration over recursion here. What do you think?

cmds/clients/contestcli-http/main.go
```diff
@@ -149,12 +183,40 @@ func run(verb string) error {
...
 	// so it can be piped to other tools if desired.
-	fmt.Println(string(indentedJSON))
-	return nil
+	return string(indentedJSON), nil
 }
+ 
+func wait(params url.Values) (string, error) {
+	
+	var finished bool
+	time.Sleep(jobWaitPoll)
```
insomniacslk, line 198: the sleep could go at the end, so that long intervals would not delay the first check too much, and it's probably desirable to have the first check shortly after the job submission to know before 10 seconds whether the job failed immediately or not

cmds/clients/contestcli-http/main.go
```diff
@@ -149,12 +183,40 @@ func run(verb string) error {
...
+	}
+	
+	var parsedResp map[string]interface{}
+	if err := json.Unmarshal([]byte(resp), &parsedResp); err != nil {
+		return "", fmt.Errorf("cannot decode json respnse: %v", err)
+	}
+	jobState := parsedResp["Data"].(map[string]interface{})["Status"].(map[string]interface{})["State"]
+
+	for _, eventName := range jobmanager.JobCompletionEvents {
+		if fmt.Sprintf("%v", jobState) == fmt.Sprintf("%v", eventName) {
```
insomniacslk, line 212: `event.Name` is aliased on `string`, so you can just use a type conversion, to eliminate the overhead of a function call (I believe that Sprintf is not optimized out here, while the type conversion should be done at compile time):
```
if string(jobState) == string(eventName) {
```

cmds/clients/contestcli-http/main.go
```diff
@@ -149,12 +183,40 @@ func run(verb string) error {
...
+	time.Sleep(jobWaitPoll)
+	
+	resp, err := request("status", params)
+	if err != nil {
+		return "", err
+	}
+	
+	var parsedResp map[string]interface{}
+	if err := json.Unmarshal([]byte(resp), &parsedResp); err != nil {
+		return "", fmt.Errorf("cannot decode json respnse: %v", err)
```
insomniacslk, line 207: In this project we have agreed to use `%w` instead of `%v` for error objects, since after Go 1.13 this allows wrapping errors without losing the original types.

Also typo: "respnse" -> "response"

cmds/clients/contestcli-http/main.go
```diff
@@ -149,12 +183,40 @@ func run(verb string) error {
...
+	}
+	jobState := parsedResp["Data"].(map[string]interface{})["Status"].(map[string]interface{})["State"]
```
insomniacslk, line 209: this double-runtime-cast is quite hacky, I think the right approach would be to return typed structures from `request()` and get the compile-time guarantee that those fields are the right ones

cmds/clients/contestcli-http/main.go
```diff
@@ -77,26 +81,56 @@ func run(verb string) error {
...
 		params.Add("jobDesc", string(jobDesc))
+		resp, err := request(verb, params)
+		if err != nil {
+			return err
+		}
+		fmt.Println(resp)
+		// if wait flag set, keep polling for status till job is completed
+		if *flagWait {
+			fmt.Fprintf(os.Stderr, "\nWaiting for job to complete...")
+			var parsedResp map[string]interface{}
```
insomniacslk, line 92: this should also be a typed structure as I suggested below, after changing the signature of `request()`, so there's no need for the typecasting dance a few lines below

cmds/clients/contestcli-http/main.go
```diff
@@ -77,26 +81,56 @@ func run(verb string) error {
...
 		}
 		params.Add("jobDesc", string(jobDesc))
+		resp, err := request(verb, params)
+		if err != nil {
+			return err
+		}
+		fmt.Println(resp)
+		// if wait flag set, keep polling for status till job is completed
+		if *flagWait {
+			fmt.Fprintf(os.Stderr, "\nWaiting for job to complete...")
```
insomniacslk, line 91: this breaks the assumption that this program only prints JSON output. I understand that `-wait` implies that we cannot print a single JSON document, but maybe this could print one-line JSON documents instead?

cmds/clients/contestcli-http/main.go
```diff
@@ -77,26 +81,56 @@ func run(verb string) error {
...
+			var parsedResp map[string]interface{}
+			if err := json.Unmarshal([]byte(resp), &parsedResp); err != nil {
+				return fmt.Errorf("cannot decode json respnse: %v", err)
+			}
+			jobID := parsedResp["Data"].(map[string]interface{})["JobID"]
+			params.Set("jobID", fmt.Sprintf("%v", jobID))
```
insomniacslk, line 97: nit: use `strconv.Itoa` instead of `fmt.Sprintf`

cmds/clients/contestcli-http/main.go
ahmomo: I think printing one line JSON is still not going to be consistent with the rest of the operations. Also changing the output of all other operations will be disruptive and beyond the purpose of a confined change to add a flag. Maybe for -wait we can skip printing the start response? Or we can just highlight this difference in the flag help.

cmds/clients/contestcli-http/main.go
ahmomo: Done

cmds/clients/contestcli-http/main.go
ahmomo: Done

cmds/clients/contestcli-http/main.go
ahmomo: Done

cmds/clients/contestcli-http/main.go
ahmomo: Done

cmds/clients/contestcli-http/main.go
ahmomo: Done

cmds/clients/contestcli-http/main.go
ahmomo: I'm leaving %v since this is what is used for the rest of the errors in the file. This can be refactored for all of them in another PR maybe.

cmds/clients/contestcli-http/main.go
ahmomo: Done, but the down side of this is an extra useless call for the tests that I'm doing since the jobs don't complete that fast. However, I'm not familiar with the system so maybe some failures can be fast enough to make use of that.

cmds/clients/contestcli-http/main.go
ahmomo: Done

cmds/clients/contestcli-http/main.go
insomniacslk: I'd say it's better to have one more call rather than having to wait a possibly too long time before the first check. Otherwise, the first check could have a short delay, e.g. 1s-3s

cmds/clients/contestcli-http/main.go
insomniacslk: OK, I won't insist. Our agreement was to avoid refactor PRs of this type, and just slowly migrate things, but this is acceptable

cmds/clients/contestcli-http/main.go
insomniacslk: highlighting the difference in the help sounds good

insomniacslk:
Thanks for updating your diff! I will review it again in a few hours as I have a wall of meetings now

cmds/clients/contestcli-http/main.go
```diff
@@ -149,12 +209,34 @@ func run(verb string) error {
...
+		}
+		jobState := parsedResp.Data.Status.State
+
+		for _, eventName := range jobmanager.JobCompletionEvents {
+			if string(jobState) == string(eventName) {
+				return resp, nil
+			}
+		}
+		time.Sleep(jobWaitPoll)
+		continue
```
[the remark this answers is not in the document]
ahmomo: You are right, removing this.

cmds/clients/contestcli-http/main.go
```diff
@@ -149,12 +209,34 @@ func run(verb string) error {
...
+			return "", err
+		}
+
+		var parsedResp statusResponse
+		if err := json.Unmarshal([]byte(resp), &parsedResp); err != nil {
+			return "", fmt.Errorf("cannot decode json response: %v", err)
+		}
+		jobState := parsedResp.Data.Status.State
+
+		for _, eventName := range jobmanager.JobCompletionEvents {
```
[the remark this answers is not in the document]
ahmomo: I put a comment there.

cmds/clients/contestcli-http/main.go
```diff
@@ -77,26 +107,56 @@ func run(verb string) error {
...
+		if err != nil {
+			return err
+		}
+		fmt.Println(resp)
 	case "version":
 		// no params for protocol version
 	default:
 		return fmt.Errorf("invalid verb: '%s'", verb)
 	}
+	return nil
```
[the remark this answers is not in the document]
ahmomo: This is the return value for error, the compiler requires it.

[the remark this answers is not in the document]
xaionaro: Whoops, sorry. I thought it is `main()` :D. Misread.

cmds/clients/contestcli-http/main.go
```diff
@@ -31,13 +34,40 @@ import (
...
+	Data struct {
+		JobID int
+	}
+}
+
+type statusResponse struct {
+	ServerID string
+	Type string
+	Error string
+	Data struct {
```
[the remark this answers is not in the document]
ahmomo: I totally agree. Will use the Data types for and later we can figure a better way to use Response directly. 

@xaionaro: many thanks for helping me out with this!

cmds/clients/contestcli-http/main.go
```diff
@@ -149,12 +209,34 @@ func run(verb string) error {
...
+			return "", fmt.Errorf("cannot decode json response: %v", err)
+		}
+		jobState := parsedResp.Data.Status.State
+
+		for _, eventName := range jobmanager.JobCompletionEvents {
+			if string(jobState) == string(eventName) {
+				return resp, nil
+			}
+		}
+		time.Sleep(jobWaitPoll)
```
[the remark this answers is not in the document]
ahmomo: will pass on this for now due to my lack of any Go knowledge ;)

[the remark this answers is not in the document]
insomniacslk: That's ok, can you add a TODO to keep track of switching to a ticker?

xaionaro:
@ahmomo : please let us know, when/if the PR will be ready for review :)

cmds/clients/contestcli-http/main.go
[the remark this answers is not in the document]
ahmomo: Done

cmds/clients/contestcli-http/main.go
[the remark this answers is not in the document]
ahmomo: Got rid of the struct definitions, thanks again @xaionaro !

ahmomo:
I have covered all comments for now. Please feel free to review this.

cmds/clients/contestcli-http/main.go
```diff
@@ -149,12 +188,36 @@ func run(verb string) error {
...
+		if err != nil {
+			return "", err
+		}
+
+		parsedData := &api.ResponseDataStatus{}
+		parsedResp := &httplistener.HTTPAPIResponse{Data: parsedData}
+		if err := json.Unmarshal([]byte(resp), parsedResp); err != nil {
+			return "", fmt.Errorf("cannot decode json response: %v", err)
+		}
+		jobState := parsedData.Status.State
```
xaionaro, line 213: Sorry, forgot to mention this earlier:

Sometimes Status may be `nil` ([it happens in `api.Status()`](https://github.com/facebookincubator/contest/blob/ebae6008ef2a5154929cf252ccdb9ff1e24b01aa/pkg/api/api.go#L201)). So this line will cause `panic` (since there's no way to dereference `nil` to access field `State`).

And therefore it makes sense to add before this line an `if`:

```go
if parsedResp.Error != nil {
    return "", fmt.Errorf("server responded with an error: %s", *parsedResp.Error)
}
```

ahmomo: Good catch.  Added  this as well.

xaionaro:
@ahmomo : look good. Thank you for the contribution!

PR #89, opened 2020-04-09
Change lock timeout from 10s to 1m

insomniacslk:
Changed to a more reasonable value. Side benefit: logs are a lot less
spammy.

### DIFF
--- a/plugins/targetlocker/inmemory/inmemory.go
+++ b/plugins/targetlocker/inmemory/inmemory.go
@@ -21,7 +21,7 @@ import (
 // Name is the name used to look this plugin up.
 var Name = "InMemory"
 
-var log = logging.GetLogger("teststeps/" + strings.ToLower(Name))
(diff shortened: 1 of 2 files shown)


plugins/targetlocker/inmemory/inmemory.go
```diff
@@ -21,7 +21,7 @@ import (
 // Name is the name used to look this plugin up.
 var Name = "InMemory"
 
-var log = logging.GetLogger("teststeps/" + strings.ToLower(Name))
+var log = logging.GetLogger("targelocker/" + strings.ToLower(Name))
```
tfg13, line 24:
```suggestion
var log = logging.GetLogger("targetlocker/" + strings.ToLower(Name))
```

insomniacslk: fat fingers :D

PR #93, opened 2020-04-24
Make EndTime optional

insomniacslk:
Since we removed the Finished flag, when a job has not ended its EndTime
is the zero-value of time.Time, and it is confusing. With this change I
am making EndTime optional, so it will result as null in the status
instead of "0001-01-01T00:00:00Z".

Test plan:
* run a job, and before it finishes, get its status
* EndTime is null, not "0001-01-01T00:00:00Z"
```
$ go run . status 16 2> /dev/null | jq '.Data.Status.EndTime'
null
```
* after the job is over, EndTime has the right value
```
$ go run . status 16 2> /dev/null | jq '.Data.Status.EndTime'
2020-04-24T13:12:55Z
```

### DIFF
--- a/pkg/job/status.go
+++ b/pkg/job/status.go
@@ -82,9 +82,8 @@ type Status struct {
 	// started yet"
 	StartTime time.Time
 
-	// EndTime indicates when the job started. This value should be ignored if
-	// `Finished` is false.
-	EndTime time.Time
+	// EndTime indicates when the job ended.
+	EndTime *time.Time
 
 	// RunStatuses represents the status of the current run of the job
 	RunStatus RunStatus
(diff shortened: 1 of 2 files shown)


pkg/job/status.go
```diff
@@ -82,9 +82,8 @@ type Status struct {
 	// started yet"
 	StartTime time.Time
 
-	// EndTime indicates when the job started. This value should be ignored if
-	// `Finished` is false.
-	EndTime time.Time
+	// EndTime indicates when the job started.
```
tfg13, line 85:
```suggestion
	// EndTime indicates when the job ended.
```

insomniacslk: wow, that has been there for quite a while :D

PR #96, opened 2020-04-27
Simplify API for Job information storage and retrieval

marcoguerri:
There is an additional layer between the storage interface
and the interface the job manager uses to access job information.
This additional layer matches the events API, but it's not really
necessary for job information. The additional layer for events
API is necessary to introduce validation of some parameters
(e.g. the query) which should not be the plugin's responsibility.
But the JobStorage API does not require any further logic, so the
additional abstraction can be removed.


marcoguerri:
This is a simplification that I wanted to do for a while. The job manager so far has been using an `Emit/Retrieve` interface to job storage (defined in the job package) that was in turn using the implementation of the `Storage` interface (defined in the storage package). However, the `Emit/Retrieve` interface serves basically no purpose, it just replicates the architecture of the events interfaces, which however are responsible for additional validation. So, this PR makes the `JobManager` use directly part of the `Storage` interface (defined in storage package) and deletes several interfaces/implementations that have no real reason to exist.

marcoguerri:
Forgot to amend with the latest fixes, so build is failing :) Fixing now.

marcoguerri:
The testing is limited to integration tests. Feel free to request more thorough testing.

marcoguerri:
Yes, that's correct. Nothing has changed for `Events`, the `Emitter/Fetcher` interfaces are still present, as they are at a higher level of abstraction and are still necessary. All the job operations however now happen directly via `storage.Storage` and all the intermediate interfaces have been removed.

PR #103, opened 2020-05-12
bugfix(storage/memory): Filter events by RunID

xaionaro:
These events are used while generating final reports.

Without this additional condition-filter events from all runs are returned for report of each run. So if there were two runs, then there will be 2 events for each target on every single run in final reports.

The patch fixes this problem.

# Test Plan

## was 

```
--- FAIL: TestMemory_GetTestEvents (0.00s)
    memory_test.go:55: 
                Error Trace:    memory_test.go:55
                Error:          "[{2020-05-13 13:47:33.836198122 +0000 UTC m=+0.000624434 %!s(*testevent.Header=&{1 2 3 4}) %!s(*testevent.Data=&{ <nil> <nil>})} {2020-05-13 13:47:33.836200697 +0000 UTC m=+0.000627005 %!s(*testevent.Header=&{1 5 3 4}) %!s(*testevent.Data=&{ <nil> <nil>})}]" should have 1 item(s), but has 2
                Test:           TestMemory_GetTestEvents
FAIL
FAIL    github.com/facebookincubator/contest/plugins/storage/memory     0.002s
```

## became

```
ok      github.com/facebookincubator/contest/plugins/storage/memory     0.002s
```

### DIFF
--- a/plugins/storage/memory/memory.go
+++ b/plugins/storage/memory/memory.go
@@ -64,6 +64,13 @@ func eventJobMatch(queryJobID types.JobID, jobID types.JobID) bool {
 	return true
 }
 
+func eventRunMatch(queryRunID, runID types.RunID) bool {
+	if queryRunID != 0 && runID != queryRunID {
+		return false
+	}
+	return true
+}
+
 func eventNameMatch(queryEventNames []event.Name, eventName event.Name) bool {
 	if len(queryEventNames) == 0 {
 		// If no criteria was specified for matching the name of the event,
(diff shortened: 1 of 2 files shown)


xaionaro:
> The code is fine, but do we agree this is a problem?

Well, I see in the code that there should be filtering by `RunID`:
```
	targetAcquiredEvents, err := jr.testEvManager.Fetch(
		testevent.QueryJobID(coordinates.JobID),
		testevent.QueryRunID(coordinates.RunID),
		testevent.QueryTestName(coordinates.TestName),
		testevent.QueryEventName(target.EventTargetAcquired),
	)
```
Also this patch fixed @insomniacslk 's reported plugin for a multi-run job (in my tests).

marcoguerri:
Mh... this seems to be definitely something missing. Was it me who said that this was expected? :O

PR #106, opened 2020-05-27
[test_runner] Refactor, prepare for cleanup pipeline

marcoguerri:
The logic of the test runner is not modular enough to accomodate the
need for implementing a cleanup pipeline. This revision refactors the
test runner so that the concept of pipeline can be re-used. The eventual
goal is to deploy two pipelines, one for testing and one for cleanup.

### DIFF
--- a/cmds/clients/contestcli-http/start.json
+++ b/cmds/clients/contestcli-http/start.json
@@ -1,6 +1,6 @@
 {
     "JobName": "test job",
-    "Runs": 3,
+    "Runs": 1,
     "RunInterval": "5s",
     "Tags": ["test", "csv"],
     "TestDescriptors": [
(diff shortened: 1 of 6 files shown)


marcoguerri:
This PR turned out quite complex to follow, as it basically touches the whole test runner. The logic has not changed, but it was split into components that can be better re-used.

* `pipeline` is now a standalone object, that can be re-used to implement multiple pipelines. This is going to be used to implement the cleanup pipeline. 
* The channel writing operations are now encapsulated via a `targetWriter` object

Let me know what your thoughts are about reviewing this. I believe that being only a refactor, integration tests would be enough to validate that there was no regression. I have few more improvements coming that will simplify the logic, and those will be more approachable.

marcoguerri:
This was actually reviewed and pushed as part of #107.

PR #108, opened 2020-06-04
Expose error in status for failed job states

insomniacslk:
This change exposes the error message in status requests when a job ends
with JobStateFailed. This makes troubleshooting easier, without having
to access the database.

### DIFF
--- a/pkg/jobmanager/event.go
+++ b/pkg/jobmanager/event.go
@@ -28,7 +28,7 @@ var EventJobCancelled = event.Name("JobStateCancelled")
 // EventJobCancellationFailed indicates that the cancellation was not completed correctly
 var EventJobCancellationFailed = event.Name("JobStateCancellationFailed")
 
-// JobCompletionEvents gather all event that mark the end of a job
+// JobCompletionEvents gathers all event names that mark the end of a job
 var JobCompletionEvents = []event.Name{
 	EventJobCompleted,
 	EventJobFailed,
@@ -36,7 +36,7 @@ var JobCompletionEvents = []event.Name{
 	EventJobCancellationFailed,
 }
 
-// JobStateEvents gather all event names which track the state of a job
+// JobStateEvents gathers all event names which track the state of a job
 var JobStateEvents = []event.Name{
 	EventJobStarted,
 	EventJobCompleted,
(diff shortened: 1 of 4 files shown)


insomniacslk:
# Testing

Executed a `cmd` step that succeeds in one case, and that fails (with a plugin error, not target error) in another.

## With a job that succeeds

Executed normally, job succeeded as expected, and no error message
```
...
        "TestName": "Literal test",
        "TestStepLabel": "some label",
        "TestStepName": "Cmd"
       }
      ]
     }
    ]
   },
   "StartTime": "2020-06-04T09:34:39Z",
   "State": "JobStateCompleted",
   "StateErrMsg": ""
  }
 },
 "Error": null
}
```

## With a job that fails

Modified the plugin to return `errors.New("injected failure"). The error surfaces as expected:
```
...
   "StartTime": "2020-06-04T09:34:09Z",
   "State": "JobStateFailed",
   "StateErrMsg": "injected failure"
  }
 },
 "Error": null
}
```

pkg/jobmanager/event.go
```diff
@@ -28,15 +28,15 @@ var EventJobCancelled = event.Name("JobStateCancelled")
 // EventJobCancellationFailed indicates that the cancellation was not completed correctly
 var EventJobCancellationFailed = event.Name("JobStateCancellationFailed")
 
-// JobCompletionEvents gather all event that mark the end of a job
+// JobCompletionEvents gathers all event that mark the end of a job
```
tfg13, line 31:
```suggestion
// JobCompletionEvents gathers all events that mark the end of a job
```

pkg/jobmanager/jobmanager.go
```diff
@@ -31,8 +31,8 @@ var log = logging.GetLogger("pkg/jobmanager")
 
 var cancellationTimeout = 60 * time.Second
 
-// errorPayload represents the payload carried by a failure event (e.g. JobStateFailed, JobStateCancelled, etc.)
-type errorPayload struct {
+// ErrorEventPayload represents the payload carried by a failure event (e.g. JobStateFailed, JobStateCancelled, etc.)
+type ErrorEventPayload struct {
 	Err string
```
xaionaro, line 36: Personally I would preserve type error here and solve the JSON-marshaling problem using `MarshalJSON` method :)

PR #110, opened 2020-06-04
[test_runner] Split pipeline and router in different files

marcoguerri:
Pipeline and routing logic are now split into different files.
Unit tests for the routing logic have been added.


marcoguerri:
Depends on #109

xaionaro:
It's very difficult to review, when all commits are in one PR. It would've been easier if the PR would be to a branch where only the last commit is absent (so it could be reviewed alone).

pkg/logging/logger.go
```diff
@@ -21,6 +21,24 @@ func GetLogger(prefix string) *logrus.Entry {
...
+}
+
+// AddFields adds multiple fields to an existing logrus.Entry
+func AddFields(e *logrus.Entry, fields map[string]interface{}) *logrus.Entry {
+	l := e
+	for k, v := range fields {
+		l = AddField(l, k, v)
+	}
+	return l
+}
```
xaionaro, line 40: As it was already discussed before, I believe builtin `logrus` functions is a better solution.

pkg/runner/test_runner_route_test.go
```diff
@@ -0,0 +1,403 @@
...
+		TestName:      "TestRunnerSuite",
+		TestStepLabel: "TestRunnerSuite",
+	}
+	ev := storage.NewTestEventEmitterFetcher(header)
+
+	timeouts := TestRunnerTimeouts{
+		StepInjectTimeout:   30 * time.Second,
+		MessageTimeout:      5 * time.Second,
+		ShutdownTimeout:     1 * time.Second,
+		StepShutdownTimeout: 1 * time.Second,
```
xaionaro, line 114: Couldn't so short timeout cause a false-positive problem in tests? (because of swapping on the testing node, for example). May be we should make all this timeouts infinite (and also remove `time.After` below)?

pkg/runner/test_runner_route_test.go
```diff
@@ -0,0 +1,403 @@
...
+	suite.router = newRouter(log, bundle, suite.routingChannels, ev, timeouts)
+}
+
+func (suite *TestRunnerSuite) TestRouteInRoutesAllTargets() {
+
+	// test that all targets are routed to step input channel without delay, in order
+	targets := []*target.Target{
+		&target.Target{Name: "host001", ID: "001", FQDN: "host001.facebook.com"},
+		&target.Target{Name: "host002", ID: "002", FQDN: "host002.facebook.com"},
+		&target.Target{Name: "host003", ID: "003", FQDN: "host003.facebook.com"},
```
xaionaro, line 126: Unnecessary `&target.Target`.

pkg/runner/test_runner_route_test.go
```diff
@@ -0,0 +1,403 @@
...
+					} else {
+						stepInResult <- nil
+					}
+					return
+				}
+				if numTargets+1 > len(targets) {
+					stepInResult <- fmt.Errorf("more targets returned than injected")
+					return
+				}
+				if t.Name != targets[numTargets].Name || t.ID != targets[numTargets].ID {
```
xaionaro, line 170: Should it be the same pointer? Why we compare fields (which could change in future) if we can compare pointers?

xaionaro:
I also skipped some comments which were already expressed previously. I had a feeling that I already reviewed some of that changes somewhere.

xaionaro:
Could you rebase this branch?

marcoguerri:
Just finished, it's not super trivial, there were several changes that I added to #109  and I shuffled logic around files and I need to be careful not to lose stuff. I rebased, and now going through the comments.

xaionaro:
The history was lost: I file was deleted and created instead of renaming and modifying :(

marcoguerri:
Now to figure out why codecov is failing the run for a slight regression in coverage....
```
codecov/project — 65.22% (+-0.02%) compared to 82708b5
```
Never seen this before...

marcoguerri:
Added a `.codecov.yml` config to allow a 20% drop in coverage:
```
.codecov.yml 
coverage:
  status:
    project:
      default:
        target: auto
        threshold: 20%
```
Let's see.

marcoguerri:
Ok looks fine now, I will leave the configuration as is and we can tweak it later if needed.

pkg/runner/test_runner.go
```diff
@@ -220,61 +225,51 @@ func (p *pipeline) route(terminateRoute <-chan struct{}, bundle test.TestStepBun
...
 			log.Debugf("received injection result for %v", injectionResult.target)
-			ingressTarget[pendingTarget] = time.Now()
 			pendingTarget = nil
 			if injectionResult.err != nil {
 				err = fmt.Errorf("routing failed while injecting target %+v into %s", injectionResult.err, stepLabel)
 				targetInErrEv := testevent.Data{EventName: target.EventTargetInErr, Target: injectionResult.target}
 				if err := ev.Emit(targetInErrEv); err != nil {
 					log.Warningf("could not emit %v event for target: %+v", targetInErrEv, *injectionResult.target)
 				}
-				break
```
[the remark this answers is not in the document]
marcoguerri: This is outdated now, but if I remember correctly, `break` was being replaced by an `else` statement. I think it's clear that you prefer the former, while I tried to go into the latter direction. The logic is now:
```
                case injectionResult := <-injectResultCh:
                        log.Debugf("received injection result for %v", injectionResult.target)
                        routeInProgress = false 
                        if injectionResult.err != nil {
                                err = fmt.Errorf("routing failed while injecting target %+v into %s", injectionResult.target, stepLabel)
                                targetInErrEv := testevent.Data{EventName: target.EventTargetInErr, Target: injectionResult.target}
                                if err := r.ev.Emit(targetInErrEv); err != nil {
                                        log.Warningf("could not emit %v event for target %+v: %v", targetInErrEv, *injectionResult.target, err)
                                }
                        } else {
                                targetInEv := testevent.Data{EventName: target.EventTargetIn, Target: injectionResult.target}
                                if err := r.ev.Emit(targetInEv); err != nil {
                                        log.Warningf("could not emit %v event for Target: %+v", targetInEv, *injectionResult.target)
                                }
                        }
```

which looks correct to me.

pkg/runner/test_runner.go
```diff
@@ -220,61 +225,51 @@ func (p *pipeline) route(terminateRoute <-chan struct{}, bundle test.TestStepBun
...
-				// no more Targets will come through. Block reading from this channel
 				log.Debugf("routing input channel closed")
 				tRouteIn = nil
-			} else {
-				// Buffer the target and check if there is already an injection in progress.
-				// If so, pending targets will be dequeued only at the next result available
-				// on `injectResultCh`.
+				if position == 0 {
+					log.Debugf("setting in targets: %d", len(ingressTarget))
+					atomic.StoreUint64(&p.numIngress, numIngress)
```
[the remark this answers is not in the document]
marcoguerri: I am not sure I follow this: the idea here is that the first step of the pipeline should keep track of the number of targets injected, which at the moment it's done in a racy way via this `numIngress` (which is being replaced in a later PR by a channel). However the check for `position==0` will still be there, because the first injection is special and needs to do this additional accounting.

pkg/runner/test_runner.go
```diff
@@ -220,61 +225,51 @@ func (p *pipeline) route(terminateRoute <-chan struct{}, bundle test.TestStepBun
...
+				}
+			}
+			if t != nil {
```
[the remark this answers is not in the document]
marcoguerri: Also this is outdated :( I have tried to rebuild the history based on #109 , but it seems this control path was already different before #109 , and I don't recall exactly the structure of the code path at the time of commenting :( I'll leave this aside, sorry for the confusion.

pkg/runner/test_runner.go
```diff
@@ -319,25 +313,38 @@ func (p *pipeline) route(terminateRoute <-chan struct{}, bundle test.TestStepBun
...
-			break
+		if pendingTarget == nil {
+			if targets.Len() > 0 {
+				// More targets available to forward to the test step
+				pendingTarget = targets.Back().Value.(*target.Target)
+				ingressTarget[pendingTarget] = time.Now()
+				targets.Remove(targets.Back())
+				injectionWg.Add(1)
+				log.Debugf("writing target %v into test step", pendingTarget)
+				go targetWriter.writeTargetWithResult(terminate, pendingTarget, injectionChannels, &injectionWg)
```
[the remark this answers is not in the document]
marcoguerri: Since this is a pipelined design, the requirement was that the targets would move asynchronously and there wouldn't be any sync point where all of them need to be available before starting a plugin run. Whether all the targets should be collected, or the test can start on a per-target basis, or any other aggregation should happen, was left to decide to the plugin. Hence the targets flowing through channels.

pkg/runner/test_runner.go
```diff
@@ -575,22 +584,32 @@ func (p *pipeline) waitTargets(terminate <-chan struct{}, ch *pipelineCtrlCh, co
...
 		}
 
 		select {
+		case target, ok := <-outChannel:
+			if !ok {
+				outChannel = nil
+				log.Debugf("pipeline output channel was closed, no more targets will come through")
+				break
+			}
+			if target != nil {
```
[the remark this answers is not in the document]
marcoguerri: `target == nil && !ok` is the configuration with closed channel and zero value coming out of the channel. The `!=nil` check is not present anymore: there should never be a nil value coming out of the channel. I am thinking that an assertion would be nice to have.

pkg/logging/logger.go
marcoguerri: Yeah, sorry for déjà vu. It's totally fair to review one PR at at time and review the other one once the rebase has happened.

pkg/runner/test_runner_route_test.go
marcoguerri: Ok, so this is a long standing problem. What makes two targets equal? So that at any point in time in a test there will not be two targets that are considered equal? There should be only one place where this is defined, e.g. `t1.Equal(t2)`, and in my opinion pointer comparison is not the right approach. However, pointer comparison is used basically all over the place in the runner, so I'll move to pointer comparison for consistency, but I'd love if we moved soon to a more robust mechanism.

xaionaro: Yeah, OK. This is a subject for discussion what internal guarantees we provide. And before that your approach is safer, so let's leave it as is, agreed.

pkg/runner/test_runner_route_test.go
marcoguerri: I consider some of these timeouts important, and the initial idea was not to overcomplicate the design by allowing per step/test timeouts. I'll give some background on those timeouts:
* MessageTimeout: is the timeout for communication within ConTest framwork. At any point in time, it should be guaranteed that ConTest is listening for a message on the receiving side. If a sender hangs, it's the indication of a bug. If this timeout fires, the framework panics.
* ShutdownTimeout: this is maximum time allowed for the pipeline to terminate after a cancellation signal. Also here, it's expected that all Steps terminate after a cancellation signal is delivered. If that doesn't happen, we want to flag the Steps as non-compliant
* StepShutdownTimeout: this is a more granular timeout when waiting for test steps when all targets have terminated. Also here, we expect the steps to terminate within a reasonable timeout.


EDIT:
I see you are specifically referring to the usage in test environment. I think for the reason above, these timeouts should be there to clearly fail the test. Possibly higher to prevent as much as possible false-positives, but I don't think they should be infinite. I'll increase by a factor 5.

pkg/runner/test_runner_route_test.go
marcoguerri: Just for the record, I reverted to pointer comparison for consistency. I am in favor of establishing the single mechanism mentioned above, which we agree provides better guarantees, but I'd rather be consistent here with what the rest of the code (and actually the pipeline which is being unit test here, does).

pkg/runner/test_runner.go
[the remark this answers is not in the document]
xaionaro: Yeah, I understand the purpose of `position==0`, but I thought to apply another approach (discussed previously). But you explained that it won't work and moreover you said this logic is reworked in further PRs, so I just "resolve" this discussion thread :)

pkg/runner/test_runner.go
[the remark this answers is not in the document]
xaionaro: Yeah, looks correct, agreed.

pkg/runner/test_runner.go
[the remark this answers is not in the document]
xaionaro: Yeah, let's not waste time :)

PR #113, opened 2020-06-11
feat(targetmanager)!: EventEmitter for TargetManager

xaionaro:
## Initial problem:
It is required to display information about why `TargetManager` acquired the targets it acquired, therefore it is required to emit events right from a `TargetManager`.

## Implementation:
To do not add an extra entity for such case it is reached a temporary agreement to use testevents with empty `TestStepLabel` for `TargetManager` events. However to be able to query such events it was required to do modify `testevent.Query`.

## BREAKING CHANGE:
Interface `TargetManager` was changed and it was moved to `github.com/facebookincubator/contest/pkg/targetmanager`.

## Post scriptum

I understand this is an arguable change. Let's iterate if you feel like we should not do it the way it is done here.


insomniacslk:
For clarity: as mentioned in the description this is a temporary, retrocompatible change until https://github.com/facebookincubator/contest/issues/123 is done

pkg/event/testevent/test.go
```diff
@@ -51,7 +51,7 @@ type Query struct {
 	event.Query
 	RunID         types.RunID
 	TestName      string
-	TestStepLabel string
+	TestStepLabel *string
```
xaionaro, line 54: We need to be able to search by empty string, so we will use `nil`-value as a zero-value (instead of `""`).

pkg/event/internal/querytools/apply_query_field.go
```diff
@@ -13,13 +13,25 @@ import (
...
 func ApplyQueryField(fieldPtr interface{}, queryField event.QueryField) error {
-	if reflecttools.IsZero(queryField) {
-		return event.ErrQueryFieldHasZeroValue{QueryField: queryField}
-	}
 	field := reflect.ValueOf(fieldPtr).Elem()
 	if !reflecttools.IsZero(field.Interface()) {
 		return event.ErrQueryFieldIsAlreadySet{FieldValue: field.Interface(), QueryField: queryField}
 	}
+
+	if field.Kind() == reflect.Ptr {
```
xaionaro, line 21: See below, we need to add support of pointer-fields.

pkg/event/internal/querytools/apply_query_field.go
```diff
@@ -13,13 +13,25 @@ import (
 )
 
 func ApplyQueryField(fieldPtr interface{}, queryField event.QueryField) error {
-	if reflecttools.IsZero(queryField) {
-		return event.ErrQueryFieldHasZeroValue{QueryField: queryField}
-	}
 	field := reflect.ValueOf(fieldPtr).Elem()
```
insomniacslk, line 16: I'm no `reflect` expert, but I tried passing an empty string pointer to this and it `panic`'ed. Unless I'm missing something, is this any risk we are subject to? 

Reproduce at https://play.golang.org/p/7-8KZbbxFbG

pkg/event/internal/querytools/apply_query_field.go
```diff
@@ -13,13 +13,25 @@ import (
 )
 
 func ApplyQueryField(fieldPtr interface{}, queryField event.QueryField) error {
```
insomniacslk, line 15: I'm easily confused by the use of reflection, and I found it hard to follow this function. Can you add a docstring here to explain what happens and how?

pkg/event/internal/querytools/apply_query_field.go
xaionaro: > Reproduce at https://play.golang.org/p/7-8KZbbxFbG

Specifically in this sandbox it panics because of `Elem()` on non-pointer (and if you will replace `X.Elem()` with `reflect.Indirect(X)` then it won't panic).

In our case here:
```go
field := reflect.ValueOf(fieldPtr).Elem()
```
`fieldPtr` is always a pointer.

pkg/event/internal/querytools/apply_query_field.go
xaionaro: OK, will do.

xaionaro: I've just added few comments. Feel free to express more concerns about readability/safety.

pkg/event/internal/querytools/apply_query_field.go
insomniacslk: thanks for the explanation, just one more question: `fieldPtr` is of type `interface{}`, which is not guaranteed to be of type `Ptr` , is that correct? And if so this could still panic?

pkg/event/testevent/test.go
marcoguerri: I don't think we can specifically search by empty string, other events might be associated to empty string as well. Why can't we search by event name? Which is basically what we are already doing for `TargetAcquired`:

```
        targetAcquiredEvents, err := jr.testEvManager.Fetch(
                testevent.QueryJobID(coordinates.JobID),
                testevent.QueryRunID(coordinates.RunID),
                testevent.QueryTestName(coordinates.TestName),
                testevent.QueryEventName(target.EventTargetAcquired),
        )
```

xaionaro: Since we want to generate arbitrary events from `TargetManager`-s we cannot search by event name.

marcoguerri: Yes, I realized this only later during the day. Since eventually we want to be able to query by the source of the emission (e.g. `TargetManager`) wouldn't it be a better to establish that the event needs to have a specific name to be picked up, and then when we fix #123 , we change the query based on name to be based on source, and we don't need to change anything else? (e.g. restore the condition that specifying an empty field in the query is not valid). 

Does it really need to be arbitrary events, for what we need to do here?

pkg/event/internal/querytools/apply_query_field.go
xaionaro: > which is not guaranteed to be of type `Ptr`, is that correct?

It is guaranteed through unit-tests. Yes: it is not as reliable as compile-time check. It will be possible to introduce compile-time checks in Go2.

If you want I can unroll this generic solution to a multiple particular solutions (to eliminate dynamic typing).

pkg/event/testevent/test.go
xaionaro: > Does it really need to be arbitrary events, for what we need to do here?

I just don't like to limit a solution to a particular task instead of covering a generic task :)
But the fact it is a temporary solution makes me think that you right (and we should make that as non-invasive as possible).

> ..., and we don't need to change anything else?

Seems to be so.

---

So, should I remove the pointer?

marcoguerri: I do agree that long term you don't want to impose any constraint in terms of name (in fact, when I wrote the first comment, I did not really think through the fact that it would be the TM itself emitting the event, which I realized only via a background thread later) :) 

> I believe I should remove the pointer, right? :)
Yes, and would restore the condition that checks for zero value in the query. I believe we should as much as possible translate "this field should be zero value" into some different condition.

pkg/event/internal/querytools/apply_query_field.go
insomniacslk: no, that's good enough, thanks for the explanation

pkg/event/testevent/test.go
xaionaro: Removed the pointer and added search by event name. But I was have to introduce other hacks (to make it so-so safe): now if a `TargetManager` tries to emit an event which name is not "`TargerManager`" then `Emit()` returns an error.

pkg/job/status.go
```diff
@@ -65,6 +65,13 @@ type TestStatus struct {
 	TestCoordinates
 	TestStepStatuses []TestStepStatus
 	TargetStatuses   []TargetStatus
+
+	// TargetManagerEvents is events emitter by a TargetManager
```
xaionaro, line 69: s/emitter/emitted/

tfg13:
What is the status with this one?

Issue #114, opened 2020-06-11
Add YAML config adapter


insomniacslk:
This has been floating around in my mind for a long time - YAML is more flexible and expressive than JSON, and it would simplify the generation of more complex job descriptors. A nice way to do that in my opinion is using something similar to Caddy's config adapters, https://caddyserver.com/docs/config-adapters . At a quick look I am not sure if we can easily reuse Caddy's adapters code though - otherwise we will probably roll out our own. CC @mholt for advice on this, if you have time :)

mholt:
The YAML adapter is just a few lines - a lib to read in YAML as a generic structure and then a call to encode it as JSON.

insomniacslk:
Thank you @mholt - and you didn't see any issue in mapping any YAML to JSON in general, or that observation is limited to the way JSON and YAML are used in Caddy?

mholt:
I don't actually use YAML myself, but I haven't heard any issues reported, either. I assume that any of the advanced features of YAML would be "unrolled" into a typical K/V structure that is compatible with JSON. I'm _sure_ there is a way to break the conversion but hopefully only with crazy/convoluted uses of YAML.

insomniacslk:
good to know, thanks a lot for your input Matt!

PR #117, opened 2020-06-12
[cmds/clients/contestcli-http] Add support for YAML job descriptors

insomniacslk:
Fixes https://github.com/facebookincubator/contest/issues/114

Added adapter for YAML job descriptors, using the -yaml flag. This flag
only works with the `start` command. The YAML document is converted to
JSON, and the JSON is passed to the server in the `start` request.

Example run:
```
$ go run . -yaml start  < start-literal.yaml
Reading from stdin...
Requesting URL http://localhost:8080/start with requestor ID 'contestcli-http'
  with params:
    requestor: [contestcli-http]
    jobDesc: [{
    "JobName": "test job",
    "Reporting": {
        "FinalReporters": [
            {
                "name": "noop"
            }
        ],
        "RunReporters": [
            {
                "name": "TargetSuccess",
                "parameters": {
                    "SuccessExpression": "\u003e80%"
                }
            },
            {
                "name": "noop"
            }
        ]
    },
    "RunInterval": "3s",
    "Runs": 3,
    "Tags": [
        "test",
        "csv"
    ],
    "TestDescriptors": [
        {
            "TargetManagerAcquireParameters": {
                "Targets": [
                    {
                        "ID": "1234",
                        "Name": "example.org"
                    }
                ]
            },
            "TargetManagerName": "TargetList",
            "TargetManagerReleaseParameters": null,
            "TestFetcherFetchParameters": {
                "Steps": [
                    {
                        "label": "some label",
                        "name": "cmd",
                        "parameters": {
                            "args": [
                                "Title={{ Title .Name }}, ToUpper={{ ToUpper .Name }}"
                            ],
                            "executable": [
                                "echo"
                            ]
                        }
                    }
                ],
                "TestName": "literal test"
            },
            "TestFetcherName": "literal"
        }
    ]
}]

The server responded with status 200 OK{
 "ServerID": "barberio-xps-13-9300",
 "Type": "ResponseTypeStart",
 "Data": {
  "JobID": 57
 },
 "Error": null
}
```

### DIFF
--- a/cmds/clients/contestcli-http/main.go
+++ b/cmds/clients/contestcli-http/main.go
@@ -19,6 +19,7 @@ import (
 	"time"
 
 	"github.com/facebookincubator/contest/pkg/api"
+	"github.com/facebookincubator/contest/pkg/config"
 	"github.com/facebookincubator/contest/pkg/jobmanager"
 	"github.com/facebookincubator/contest/plugins/listeners/httplistener"
 )
@@ -42,6 +43,7 @@ var (
 	flagAddr      = flag.String("addr", "http://localhost:8080", "ConTest server [scheme://]host:port[/basepath] to connect to")
 	flagRequestor = flag.String("r", defaultRequestor, "Identifier of the requestor of the API call")
 	flagWait      = flag.Bool("wait", false, "After starting a job, wait for it to finish, and exit 0 only if it is successful")
+	flagYAML      = flag.Bool("yaml", false, "Parse job descriptor as YAML instead of JSON")
 )
 
 func main() {
@@ -67,7 +69,7 @@ func main() {
 	flag.Parse()
 	verb := flag.Arg(0)
 	if err := run(verb); err != nil {
-		fmt.Fprintf(os.Stderr, "%s", err)
+		fmt.Fprintf(os.Stderr, "%s\n", err)
 		os.Exit(1)
 	}
 }
@@ -82,9 +84,18 @@ func run(verb string) error {
 		fmt.Fprintf(os.Stderr, "Reading from stdin...\n")
 		jobDesc, err := ioutil.ReadAll(os.Stdin)
 		if err != nil {
-			return fmt.Errorf("failed to parse job descriptor: %v", err)
+			return fmt.Errorf("failed to read job descriptor: %v", err)
 		}
-		params.Add("jobDesc", string(jobDesc))
+
+		jobDescFormat := config.JobDescFormatJSON
+		if *flagYAML {
+			jobDescFormat = config.JobDescFormatYAML
+		}
+		jobDescJSON, err := config.ParseJobDescriptor(jobDesc, jobDescFormat)
+		if err != nil {
+			return fmt.Errorf("failed to parse job descriptor: %w", err)
+		}
+		params.Add("jobDesc", string(jobDescJSON))
 		resp, err := request(verb, params)
 		if err != nil {
 			return err
--- a/cmds/clients/contestcli-http/start-literal.yaml
+++ b/cmds/clients/contestcli-http/start-literal.yaml
@@ -1,0 +1,28 @@
+JobName: test job
+Runs: 3
+RunInterval: 3s
+Tags: [test, csv]
+TestDescriptors:
+    - TargetManagerName: TargetList
+      TargetManagerAcquireParameters:
+          Targets:
+              - Name: example.org
+                ID: "1234"
+      TargetManagerReleaseParameters:
+      TestFetcherName: literal
+      TestFetcherFetchParameters:
+          TestName: literal test
+          Steps:
+              - name: cmd
+                label: some label
+                parameters:
+                    executable: [echo]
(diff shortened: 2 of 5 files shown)


insomniacslk:
Forgot to attach the YAML descriptor:
```
JobName: test job
Runs: 3
RunInterval: 3s
Tags: [test, csv]
TestDescriptors:
    - TargetManagerName: TargetList
      TargetManagerAcquireParameters:
          Targets:
              - Name: example.org
                ID: "1234"
      TargetManagerReleaseParameters:
      TestFetcherName: literal
      TestFetcherFetchParameters:
          TestName: literal test
          Steps:
              - name: cmd
                label: some label
                parameters:
                    executable: [echo]
                    args: ["Title={{ Title .Name }}, ToUpper={{ ToUpper .Name }}"]
Reporting:
    RunReporters:
        - name: TargetSuccess
          parameters:
              SuccessExpression: ">80%"
        - name: noop
    FinalReporters:
        - name: noop
```

insomniacslk:
> The only thing I don't like here is that the code is isolated to `main` of a client.
> So every new client will have to duplicate this logic. Just a thought: shouldn't we support YAML on the server side, instead?

I thought about it while making this patch, and I'm not sure if it 's a good idea to support more than one communication format (JSON and YAML) through the API, I'd rather keep it simple to the server (i.e. just support one thing, JSON) while leaving the translation to the client.

Internally we have a client library with some infra-specific (i.e. useless externally) logic. I'm wondering if there's anything in that client library that can be useful publicly too, in which case we should publish it, and the YAML -> JSON translation logic could reside there. Any thoughts?

xaionaro:
I think I agree here. I just haven't thought it through, yet. Could you help me?

What I'm thinking from the bad side of this solution is:
* Total solution is no simpler, just logic moved from the server to a client.
* Locking-in the client to Go (since this logic is implemented only on Go).
* Forces ConTest consumer to understand more. A consumer writes clients and plugins. So he will have to understand this part while writing a client.
* API suggest an arbitrary string anyway, why not leverage this?

Why I think I still agree with out despite of the problems above:
* YAML is a specific case, so we should not grow server's logic because of one specific case.
* We don't want to overload documentation of the API.

In total my opinion is:
If we want to declare YAML as a core feature, then we need to move it to server. If not, then we need to keep it in client.

insomniacslk:
I'm pretty much on the same boat. The biggest thing that is holding me back from moving this to the server side is the desire to keep the server API simple and isolated from any new format we (or our users) may want to support in the future.

I agree with the arguments against having it on the client side. Just help me understand the part where the plugin writers have to understand more: using YAML should be transparent to them, because they will still receive the same data structures, and the values are formatted the same way as JSON (minus small syntax differences, that don't leak out to the plugins). What's your concern on that?

On this specific feature, while it generally hurts me to duplicate code, I'm thinking that this is so small and easy to reimplement correctly that it may be worth the duplication

xaionaro:
> What's your concern on that?

They will have to use some special package to do this conversion, while right now they are free to work with a server directly. I mean if we will start to collect all this things as a somekind of ConTest Client Framework then users will have to study this framework. May be the problem is far-fetched.

xaionaro:
Let's merge as is :)

insomniacslk:
> > What's your concern on that?
> 
> Users will have to use some special package to do this conversion, while right now they are free to work with a server directly. I mean if we will start to collect all this things as a somekind of ConTest Client Framework then users will have to study this framework (which is more difficult than flags to a request to the server). May be the problem is far-fetched.

I see what you meant now, thanks for the explanation! I agree with you, and I think I addressed this point implicitly in my reply above, when I said that I agree with you that it's better to avoid a client library at this stage, and for this feature the duplication may be acceptable. I should have been more explicit on this - do you agree that the code that converts from YAML to JSON is simple enough that this duplication in various clients is acceptable?

xaionaro:
> I should have been more explicit on this - do you agree that the code that converts from YAML to JSON is simple enough that this duplication in various clients is acceptable?

Yes. Thank you for including my weak opinion here :)

insomniacslk:
Maybe a better alternative is to move the YAML->JSON snippet to a function under `pkg/` and let clients reuse it?

xaionaro:
> Maybe a better alternative is to move the YAML->JSON snippet to a function under `pkg/` and let clients reuse it?

I believe this is a good idea, but either solution is good for me personally :)

insomniacslk:
Moved under `config`, exported where necessary. Created type `JobDescFormat` and relative constants `JobDescFormatJSON` and `JobDescFormatYAML`, and using a `switch` instead of an `if yaml` to be open to support more formats in the future

insomniacslk:
Tested with `start-literal.json` and `start-literal.yaml` (the latter with the `-yaml` flag)

insomniacslk:
@xaionaro and I have been discussing that at this point we could get rid entirely of JSON in the API, because JSON parses just fine as YAML[1]. This is a broader decision with potential ramifications, and I think we should be sure about it before doing that step. For this reason I'll merge this as it is, and track the switch to YAML in a separate issue

[1] https://eli.thegreenplace.net/2008/11/22/json-yaml-but-yaml-json

PR #118, opened 2020-06-15
[test_runner] Prepare support for cleanup steps

marcoguerri:
This PR goes into two directions:

1) Adds support for the concept of cleanup steps. The syntax to define cleanup steps is the same the one for test steps. The concept of `TestStep` has been renamed into generic `Step`, as it now can represent both a test step and a cleanup step.
2) Fixed the handling of steps descriptor, which I think has several problems:
    * The `TestDescriptors` which was added to `Request` object in my opinion in not at the right level of abstraction. `TestDescriptors` should be inside the `JobDescriptor`, which is also included in the `Request` object.
    * One of the reasons to persist `TestDescriptors` into the database is obviously because we want to be able to rebuild the status of the job, even  if the test description underneath has changed. But the PR that introduced the changes to persist `TestDescriptors` in the database, did not add support for actually re-reading those descriptors when calculating job status. So the result is that when the status is calculated, we fetch again the description from the original source (which might have changed meanwhile)


marcoguerri:
Please hold review, I am attaching more documentation to navigate through this PR.

pkg/jobmanager/job.go
```diff
@@ -0,0 +1,105 @@
...
+
+		bundleTest, bundleCleanup, err := newStepsBundles(thisTestStepsDescriptors, registry)
+		if err != nil {
+			return nil, fmt.Errorf("could not create cleanup step bundles: %v", err)
+		}
+		test := test.Test{
+			Name:                thisTestStepsDescriptors.TestName,
+			TargetManagerBundle: bundlTargetManager,
+			TestFetcherBundle:   bundleTestFetcher,
+			TestStepBundles:     bundleTest,
```
xaionaro, line 59: Sometimes we call it `StepBundle` and sometimes `StepsBundle`. I believe we should pick one to do not mislead anybody.

pkg/jobmanager/job.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// NewJobFromDescriptor creates a job object from a job descriptor
+func NewJobFromDescriptor(registry *pluginregistry.PluginRegistry, jobDescriptor *job.Descriptor) (*job.Job, error) {
+	resolver := fetcherStepsResolver{jobDescriptor: jobDescriptor, registry: registry}
+	return newJob(registry, jobDescriptor, resolver)
+}
+
+// NewJobFromExtendedDescriptor creates a job object from an extended job descriptor
+func NewJobFromExtendedDescriptor(registry *pluginregistry.PluginRegistry, jobDescriptor *job.ExtendedDescriptor) (*job.Job, error) {
```
xaionaro, line 102: I think more self-descriptive name would be `RestoreJobFromExtendedDescriptor`. Otherwise it's unclear how it will resolve steps (from scratch using the initial job descriptor or will re-use already resolved steps).

marcoguerri:
Re-basing, then addressing all comments.

pkg/test/step.go
```diff
@@ -55,57 +55,63 @@ func (t TestStepParameters) GetInt(k string) (int64, error) {
...
-	TestStepLabel string
-	Parameters    TestStepParameters
+// StepsDescriptors bundles together Test and Cleanup descriptions
+type StepsDescriptors struct {
+	TestName string
+	Test     []StepDescriptor
+	Cleanup  []StepDescriptor
+}
+
+// StepBundle contains the instantiation
```
tfg13, line 85: Instantiation of what? I don't think I have understood what a bundle is yet :D

pkg/job/job.go
```diff
@@ -14,21 +16,58 @@ import (
...
+		}
+	}
+	return nil
+}
+
+// ExtendedDescriptor is a job descriptor which has been extended with the full
+// description of the test and cleanup steps obtained from the test fetcher.
+type ExtendedDescriptor struct {
+	Descriptor
+	StepsDescriptors []test.StepsDescriptors
```
tfg13, line 58: Why is this an array of multiple`StepsDescriptors`? Doesn't one `StepsDescriptors` already contain multiple Test and Cleanup Steps?

pkg/test/step.go
tfg13: Oh is this the golang datastructure version of `StepDescriptor`?

pkg/job/job.go
xaionaro: I read this PR long time ago and can remember things wrong, but IIRC one `test.StepsDescriptors` is a set of steps for only one test (in terms of JobDescription), while `ExtendedDescriptor` contains all tests.

dimkup:
As far as I understood, StepDescriptor ([]test.StepsDescriptor) is actually, khm, part of job's state. Could it be named as <Cached|Resolved|Fetched|Prepared.. etc>StepDescriptor, just to reflect that it's an execution artefact,  not input data?

Also it would be handy to allow to read Events at least for Cleanup steps. (ie pass ev testevent.EmitterFetcher as a parameter to Run) I understand that it's pretty invasive and backward compatibility breaking change, but probably it's better to group such changes together.)

pkg/jobmanager/bundles.go
```diff
@@ -0,0 +1,97 @@
...
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to create bundle for final reporter '%s': %v", reporter.Name, err)
+		}
+		finalReporterBundles = append(finalReporterBundles, bundle)
+	}
+
+	return runReporterBundles, finalReporterBundles, nil
+}
+
+func newBundlesFromSteps(descriptors []test.StepDescriptor, registry *pluginregistry.PluginRegistry) ([]test.StepBundle, error) {
```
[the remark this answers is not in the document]
marcoguerri: Mh... the intention of this naming was to highlight that the bundles here are created from descriptors passed directly as part of the signature of the function, rather than fetched inside the function. I feel this information is lost with the naming suggested?

pkg/test/step.go
```diff
@@ -55,57 +55,63 @@ func (t TestStepParameters) GetInt(k string) (int64, error) {
...
 }
 
-// TestStepBundle bundles the selected TestStep together with its parameters as
-// specified in the Test descriptor fetched by the TestFetcher
-type TestStepBundle struct {
-	TestStep      TestStep
-	TestStepLabel string
-	Parameters    TestStepParameters
+// StepsDescriptors bundles together Test and Cleanup descriptions
+type StepsDescriptors struct {
```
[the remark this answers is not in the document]
marcoguerri: Yeah, the intention since the introduction of these terms has been to use `Descriptor` as something that results from a deserialization, and `Bundle` as a collection of instances coming from instantiating objects based on the `Descriptor`.

pkg/jobmanager/job.go
marcoguerri: You are right. As far as I can tell, the the `.*StepsBundle.*` nomenclature is used only in `newStepsBundles`, and that shoul dbe `StepBundles`. We should only have `StepBundle` or `StepBundles` (i.e. []StepBundle).

pkg/jobmanager/job.go
marcoguerri: Mh, interesting. I have the feeling that `Restore` is actually more misleading. Do you mind if I leave this as is?

xaionaro: Yes, sure.

pkg/test/step.go
marcoguerri: I often pass out while writing comments. So, a `Descriptor` object is something on which we deserialize raw JSON. From a `Descriptor`, we build a `Bundle`, which holds the instances (e.g. of the Step) that were built from the `Descriptor`. I came up with this naming, but now I am not very happy about this naming, so I am definitely open for changing this (so for all the alternatives which came to mind are worse :/)

pkg/job/job.go
marcoguerri: Yeah, what @xaionaro said, this `[]test.StepsDescriptors` is rebuilt from the serialization in the db which contains, for every `Test`, all the `StepsDescriptors`.

marcoguerri:
@dimkup , you comment seems to be going in the same direction as @xaionaro comment (with respect to 's/NewJobFromExtendedDescriptor/RestoreJobFromExtendedDescriptor`. It could actually be input data. When the job is first submitted, `[]test.StepsDescriptor` is the result of the output of the plugin. Then it is also serialized in the db and upon when resuming jobs, the fetched version will be used. In my opinion this design is relatively easy to reason around, I don't feel the strong need for a renaming. 

As for reading events, yeah, it would be probably necessary, though it needs more thinking. The idea of allowing test steps to read events was originally there, though it was reflected only on the `Resume` signature (`Resume(cancel, pause <-chan struct{}, ch StepChannels, params StepParameters, ev testevent.EmitterFetcher`), which none of the steps implement yet. This could be ported also to `Run`, but it wouldn't serve the purpose that you probably have in mind. Every emitter has metadata which allows fetching events that only the same step emitted (one of the purposes was to keep track of state for `Resume`, hence the read capabilities in the `Resume` signature). I think we did not want for a `Step` to be able to read any event, because that might lead into message passing behaviors between steps which might be abused and result in test flows very difficult to follow/debug.

marcoguerri:
We discussed that before landing this, we want to have a tool which allows for schema migration ready, and start versioning the db schema. It's going to take a while longer to get there. I am targeting next week to make it happen, as I really want to get it done with this stack of PRs). For the moment, I'll put this on hold.

marcoguerri:
Resuming and rebasing this, after #154 was merged.

marcoguerri:
Being superseded by #208

PR #119, opened 2020-06-15
[test_runner] Added support for running two pipelines back to back

marcoguerri:
This patch adds support for implementing a cleanup pipeline, which runs
cleanup steps, after the main test pipeline. One of the requirements of
ConTest has always been that it should support a "cleanup" phase, and
this adds support for it.
    
The cleanup pipeline is described in the same was test steps have been
described so far: and additional "cleanup" section in the test descriptor
is deserialized to implement a pipeline which runs the required cleanup
steps.


marcoguerri:
# Test plan

Considering a test description as follows:

```
{
    "steps": [
        {
            "name": "randecho",
            "label": "thefirstrandecho_test",
            "parameters": {
                "text": ["some very random text first_test"]
            }
        },
        {
            "name": "randecho",
            "label": "thesecondrandecho_test",
            "parameters": {
                "text": ["some very random text second_test"]
            }
        }
   ],
   "cleanup": [
        {
            "name": "randecho",
            "label": "thefirstrandecho_cleanup",
            "parameters": {
                "text": ["some very random text first_cleanup"]
            }
        }
   ]
}
```

* If a target fails `thefirstrandecho_test`, it is not going to execute `thesecondrandecho_test`, but it will be injected into the second pipeline and start executing `thefirstrandecho_cleanup`: 
   * Failed `thefirstrandecho_test`: https://gist.github.com/marcoguerri/47ce9ddb5601de0e08b45c9deaf712bb#file-02-contest-cleanup-L44
    * Started `thefirstrandecho_cleanup`: https://gist.github.com/marcoguerri/47ce9ddb5601de0e08b45c9deaf712bb#file-02-contest-cleanup-L90
* If a target succeeds in `thefirstrandecho_test`, it will normally execute `thesecondrandecho_test` and then start also going through the cleanup pipeline: 
    * Successful test steps: https://gist.github.com/marcoguerri/aade1c321f7c5814909aaae9d972fd06#file-01-contest-cleanup-L44-L61
     * Successful cleanup step: https://gist.github.com/marcoguerri/aade1c321f7c5814909aaae9d972fd06#file-01-contest-cleanup-L96

pkg/runner/test_runner.go
```diff
@@ -152,68 +153,187 @@ func newTargetWriter(log *logrus.Entry, timeouts TestRunnerTimeouts) *targetWrit
 	return &targetWriter{log: log, timeouts: timeouts}
 }
 
+func (tr *TestRunner) injectTargets(terminate <-chan struct{}, targets []*target.Target, writeChannel chan<- *target.Target, log *logrus.Entry) {
```
xaionaro, line 156: Just a thought:

```
terminate <-chan struct{}, ..., log *logrus.Entry
```
This is early stages of the problem I'm talking about here: https://fb.quip.com/PUndAbbUFp2p
Since the document might be closed for some users, I'll write TL;DR here:
Basically the thought is that all those loggers, done-contexts (`terminate` in this case), metrics, tracing and other things evolves to a set of argument to each called function through almost any execution flow. And each of them requires manual support each time (to add fields to logger, tags to metrics, recursively create a done-context and so on). So may be in long term we should consider using of smart contexts instead.

marcoguerri:
Being superseded by #208

PR #121, opened 2020-06-16
[test_runner] Simplify use of channels in step interface

marcoguerri:
NOTE: this is an API breaking change.

The interface of TestSteps so far has dictated that targets should be
forwarded to two different channels for success and failure. This makes
several parts of ConTest logic relatively complicated. In order to
simplify the framework, this braking change reduces the number out
output channels that a test step needs to handle to only one. The object
returned by the step might include error information for targets which
failed in the test step.


marcoguerri:
The relevant change to understand this PR is to be looked into the `StepChannels` struct, which changes from

```
type StepChannels struct {
    In  <-chan *target.Target
    Out chan<- *target.Target
    Err chan<- cerrors.TargetError
}
```

to 

```
type StepChannels struct {
    In  <-chan *target.Target
    Out chan<- *target.Result
}
```

where `target.Result` is the following:

```
type Result struct {
        Target *Target
        Err    error
}
```

plugins/teststeps/example/example.go
```diff
@@ -53,36 +54,38 @@ func (ts *Step) Run(cancel, pause <-chan struct{}, ch test.StepChannels, _ test.
...
-			} else {
-				select {
-				case <-cancel:
-				case <-pause:
-					return nil
-				case ch.Out <- target:
-					if err := ev.Emit(testevent.Data{EventName: FailedEvent, Target: target, Payload: nil}); err != nil {
-						return fmt.Errorf("failed to emit failed event: %v", err)
+				} else {
+					log.Infof("target %v succeded", result.Target)
```
xaionaro, line 84: **nitpicking:**

A typo: `succeded` → `succeeded`

plugins/teststeps/example/example.go
```diff
@@ -53,36 +54,38 @@ func (ts *Step) Run(cancel, pause <-chan struct{}, ch test.StepChannels, _ test.
...
-						return fmt.Errorf("failed to emit finished event: %v", err)
+				result.Err = fmt.Errorf("target failed")
+			}
+
+			select {
+			case <-cancel:
+			case <-pause:
+				return nil
+			case ch.Out <- result:
+				if result.Err != nil {
```
xaionaro, line 78: **just a thought:**

Since result is transmitter as a pointer, there a potential race condition here. We create a constraint to guarantee we will not change the result in the core. It seems to be not the best example.

Why not to emit the event before sending the result?

plugins/teststeps/example/example.go
```diff
@@ -53,36 +54,38 @@ func (ts *Step) Run(cancel, pause <-chan struct{}, ch test.StepChannels, _ test.
...
-				case <-pause:
-					return nil
-				case ch.Err <- cerrors.TargetError{Target: target, Err: fmt.Errorf("target failed")}:
-					if err := ev.Emit(testevent.Data{EventName: FinishedEvent, Target: target, Payload: nil}); err != nil {
-						return fmt.Errorf("failed to emit finished event: %v", err)
+				result.Err = fmt.Errorf("target failed")
+			}
+
+			select {
+			case <-cancel:
```
xaionaro, line 74: **warning:**

Missing `passthrough` or `return nil`.

plugins/teststeps/randecho/randecho.go
```diff
@@ -58,31 +58,26 @@ func (e Step) Name() string {
...
 			if r == 0 {
-				evData := testevent.Data{
-					EventName: event.Name("TargetSucceeded"),
-					Target:    target,
-					Payload:   nil,
-				}
-				_ = ev.Emit(evData)
-				log.Infof("Run: target %s succeeded: %s", target, params.GetOne("text"))
-				ch.Out <- target
+				eventName = event.Name("TargetSucceeded")
```
xaionaro, line 71: **nitpicking:**

Extra `event.Name(`.

marcoguerri:
Being superseded by #208

PR #126, opened 2020-07-23
[plugins/targetlocker/inmemory] several fixes

insomniacslk:
There were several bugs in the in-memory target locker:
* requests were passed by value instead of reference, so the results
  were not visible to the caller
* owner was not set and not used for lock validation

Also:
* added validation of owner and number of targets in request objects
* added many tests to make sure this doesn't happen again
* better log messages

### DIFF
--- a/plugins/targetlocker/inmemory/inmemory.go
+++ b/plugins/targetlocker/inmemory/inmemory.go
@@ -47,10 +47,23 @@ type lock struct {
 	expiresAt time.Time
 }
 
+func validateRequest(req *request) error {
+	if req == nil {
+		return fmt.Errorf("got nil request")
+	}
+	if req.owner == 0 {
+		return fmt.Errorf("owner cannot be zero")
+	}
+	if len(req.targets) == 0 {
+		return fmt.Errorf("no target specified")
+	}
+	return nil
+}
+
 // broker is the broker of locking requests, and it's the only goroutine with
 // access to the locks map, in accordance with Go's "share memory by
 // communicating" principle.
-func broker(lockRequests, unlockRequests, checkLocksRequests <-chan request, done <-chan struct{}) {
+func broker(lockRequests, unlockRequests, checkLocksRequests <-chan *request, done <-chan struct{}) {
 	locks := make(map[target.Target]lock)
 	for {
 		select {
@@ -58,7 +71,10 @@ func broker(lockRequests, unlockRequests, checkLocksRequests <-chan request, don
 			log.Debugf("Shutting down in-memory target locker")
 			return
 		case req := <-lockRequests:
-			log.Debugf("Requested to lock %d targets: %v", len(req.targets), req.targets)
+			if err := validateRequest(req); err != nil {
+				req.err <- fmt.Errorf("lock request: %w", err)
+			}
+			log.Debugf("Requested to lock %d targets for job ID %d: %v", len(req.targets), req.owner, req.targets)
 			var lockErr error
 			for _, t := range req.targets {
 				now := time.Now()
@@ -80,7 +96,7 @@ func broker(lockRequests, unlockRequests, checkLocksRequests <-chan request, don
 							l.expiresAt = time.Now().Add(req.timeout)
 							locks[*t] = l
 						} else {
-							lockErr = fmt.Errorf("target already locked: %+v", t)
+							lockErr = fmt.Errorf("lock request: target already locked: %+v", t)
 						}
 						break
 					}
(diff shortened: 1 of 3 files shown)


plugins/targetlocker/inmemory/inmemory.go
```diff
@@ -47,18 +47,34 @@ type lock struct {
 	expiresAt time.Time
 }
 
+func validateRequest(req *request) error {
```
insomniacslk, line 50: this method is called in the request broker below - every request needs valid fields

plugins/targetlocker/inmemory/inmemory.go
```diff
@@ -47,18 +47,34 @@ type lock struct {
...
+		return fmt.Errorf("no target specified")
+	}
+	return nil
+}
+
 // broker is the broker of locking requests, and it's the only goroutine with
 // access to the locks map, in accordance with Go's "share memory by
 // communicating" principle.
-func broker(lockRequests, unlockRequests, checkLocksRequests <-chan request, done <-chan struct{}) {
+func broker(lockRequests, unlockRequests, checkLocksRequests <-chan *request, done <-chan struct{}) {
```
insomniacslk, line 66: made this a pointer. The object is modified here, and without a pointer the changes are not reflected at the caller (i.e. the function that writes to the requests channels)

plugins/targetlocker/inmemory/inmemory.go
```diff
@@ -47,18 +47,34 @@ type lock struct {
...
 		select {
 		case <-done:
 			log.Debugf("Shutting down in-memory target locker")
 			return
 		case req := <-lockRequests:
-			log.Debugf("Requested to lock %d targets: %v", len(req.targets), req.targets)
+			if err := validateRequest(req); err != nil {
+				req.err <- fmt.Errorf("lock request: %w", err)
+			}
+			log.Debugf("Requested to lock %d targets for job ID %d: %v", len(req.targets), req.owner, req.targets)
```
insomniacslk, line 77: now also printing job IDs, as they are important for debugging

plugins/targetlocker/inmemory/inmemory.go
```diff
@@ -80,7 +96,7 @@ func broker(lockRequests, unlockRequests, checkLocksRequests <-chan request, don
 							l.expiresAt = time.Now().Add(req.timeout)
 							locks[*t] = l
 						} else {
-							lockErr = fmt.Errorf("target already locked: %+v", t)
+							lockErr = fmt.Errorf("lock request: target already locked: %+v", t)
```
insomniacslk, line 99: better logging: it will be easier to spot which code snipped generated it

plugins/targetlocker/inmemory/inmemory.go
```diff
@@ -95,30 +111,48 @@ func broker(lockRequests, unlockRequests, checkLocksRequests <-chan request, don
...
 		case req := <-unlockRequests:
+			if err := validateRequest(req); err != nil {
+				req.err <- fmt.Errorf("unlock request: %w", err)
+			}
 			log.Debugf("Requested to transactionally unlock %d targets: %v", len(req.targets), req.targets)
+			var unlockErr error
 			for _, t := range req.targets {
-				if _, ok := locks[*t]; ok {
-					delete(locks, *t)
+				if l, ok := locks[*t]; ok {
```
insomniacslk, line 120: previously, in unlock requests, I was just deleting the lock, if found. This is wrong when the lock is not owned by the same job ID, so now job IDs are checked

plugins/targetlocker/inmemory/inmemory.go
```diff
@@ -95,30 +111,48 @@ func broker(lockRequests, unlockRequests, checkLocksRequests <-chan request, don
...
+					}
 				} else {
-					log.Debugf("Target is not locked, but received unlock request: %+v", t)
+					// XXX should we roll back the unlock if a target fails?
+					unlockErr = fmt.Errorf("unlock request: target is not locked: %+v", t)
+					break
 				}
 			}
-			req.err <- nil
+			req.err <- unlockErr
```
insomniacslk, line 132: unlock requests always succeeded, but since we now check for lock owners, having an error is necessary

plugins/targetlocker/inmemory/inmemory.go
```diff
@@ -95,30 +111,48 @@ func broker(lockRequests, unlockRequests, checkLocksRequests <-chan request, don
...
 			notLocked := make([]*target.Target, 0)
 			for _, t := range req.targets {
 				if l, ok := locks[*t]; ok {
-					now := time.Now()
-					if now.After(l.expiresAt) {
-						// target was locked but lock expired, purge the entry
-						log.Debugf("Purged expired lock for target %+v. Lock time is %s, expiration timeout is %s", t, l.lockedAt, req.timeout)
-						delete(locks, *t)
-						notLocked = append(notLocked, t)
+					if l.owner == req.owner {
```
insomniacslk, line 142: checklocks also didn't check for the lock owner - now it returns correctly the locks that are owned by the request owner. All the rest is returned as non-locked or not-locked-by-you

plugins/targetlocker/inmemory/inmemory.go
```diff
@@ -132,70 +166,72 @@ func broker(lockRequests, unlockRequests, checkLocksRequests <-chan request, don
...
-	lockRequests, unlockRequests, checkLocksRequests chan request
+	lockRequests, unlockRequests, checkLocksRequests chan *request
 	done                                             chan struct{}
-	timeout                                          time.Duration
+	// timeout set on each lock request
+	timeout time.Duration
 }
 
-func newReq(targets []*target.Target) request {
+func newReq(jobID types.JobID, targets []*target.Target) request {
```
insomniacslk, line 177: require to pass a job ID to construct a valid request, as explained above

plugins/targetlocker/inmemory/inmemory.go
```diff
@@ -132,70 +166,72 @@ func broker(lockRequests, unlockRequests, checkLocksRequests <-chan request, don
...
+	tl.checkLocksRequests <- &req
 	if err := <-req.err; err != nil {
 		// TODO the target.Locker.CheckLocks interface has to change to return an
 		// error as well, because lock checking can fail, e.g. when using
 		// external services.
 		// Just log an error for now.
 		log.Warningf("Error when trying to unlock targets: %v", err)
 	}
-	return len(req.notLocked) != len(targets), req.locked, req.notLocked
+	return len(req.locked) == len(targets), req.locked, req.notLocked
```
insomniacslk, line 227: this check was wrong - the returned boolean means "all the targets you specified are locked by you". Adjusted to do just that

PR #130, opened 2020-08-17
config: increase lock timeout to be > target manager timeout

tfg13:
Defaults should be safe. If a target manager dynamically locks a target,
and then does some more things, it can become unlocked because the timeout
expires.

It is possible to write fancy target managers that know this and
refresh locks, but we should not require it by default.

Signed-off-by: Tobias Fleig <tfleig@fb.com>


tfg13:
Ok we discussed this and I think we found a way that is a bit more elegant, putting up a new one in a sec

PR #133, opened 2020-08-18
Add dir parameter in cmd plugin

v-thakkar:
Add dir parameter in cmd plugin so that we don't need to specify directory path in args when running commands under a different directory than default one.

Don't throw an error when this parameter is not used.

Signed-off-by: Vaishali Thakkar <vaishali.thakkar@9elements.com>

### DIFF
--- a/plugins/teststeps/cmd/cmd.go
+++ b/plugins/teststeps/cmd/cmd.go
@@ -46,12 +46,14 @@ var Events = []event.Name{
 type eventCmdStartPayload struct {
 	Path string
 	Args []string
+	Dir string
 }
 
 // Cmd is used to run arbitrary commands as test steps.
 type Cmd struct {
 	executable string
 	args       []test.Param
+	dir        string
 }
 
 // Name returns the plugin name.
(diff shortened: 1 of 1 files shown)


tfg13 (approved):
lgtm 👍

insomniacslk:
Thanks @v-thakkar , I've also tested it with this spec and works like a charm :) 
```
JobName: test job
Runs: 3
RunInterval: 3s
Tags: [test, csv]
TestDescriptors:
    - TargetManagerName: TargetList
      TargetManagerAcquireParameters:
          Targets:
              - Name: example.org
                ID: "1234"
      TargetManagerReleaseParameters:
      TestFetcherName: literal
      TestFetcherFetchParameters:
          TestName: literal test
          Steps:
              - name: cmd
                label: some label
                parameters:
                    executable: [ls]
                    dir: [/tmp]
                    args: [-la]
Reporting:
    RunReporters:
        - name: TargetSuccess
          parameters:
              SuccessExpression: ">80%"
        - name: noop
    FinalReporters:
        - name: noop
```

insomniacslk (approved):

PR #135, opened 2020-08-20
plugins/targetlocker/inmemory: Fix request validation

tfg13:
If validation fails for any reason (for example 0 targets in the request), the code still continues, but already wrote to the error channel. It later tries to write again, which won't work and locks up lock processing forever.

Signed-off-by: Tobias Fleig <tfleig@fb.com>


marcoguerri (approved):

marcoguerri:
Not strictly related: I am not familiar with this code path, but I don't see `done` channel closed anywhere in the code. Are we leaking a `broker` for every job?

tfg13:
No we only ever create one instance of the locker from the server main function, it only shuts down when the entire server shuts down

Issue #134, opened 2020-08-19
inmemory locker bugs


tfg13:
There are multiple issues with this one.

- The most severe one is probably that `RefreshLocks` will only refresh the very first lock, but still return success. This repro uses `sleep` which is not great, but reproduces it (put in `inmemory_test.go`)
```
func TestRefreshMultiple(t *testing.T) {
	tl := New(200 * time.Millisecond, 200 * time.Millisecond)
	require.NoError(t, tl.Lock(jobID, twoTargets))
	time.Sleep(100 * time.Millisecond)
	// they are not expired yet, extend both
	require.NoError(t, tl.RefreshLocks(jobID, twoTargets))
	time.Sleep(150 * time.Millisecond)
	// if they were refreshed properly, they are still valid and attempts to get them must fail
	require.Error(t, tl.Lock(otherJobID, []*target.Target{&targetOne}))
	require.Error(t, tl.Lock(otherJobID, []*target.Target{&targetTwo}))
}
```
```
=== RUN   TestRefreshMultiple
time="2020-08-19T16:47:52+01:00" level=info msg="Trying to lock 2 targets" prefix="targetlocker/inmemory"
time="2020-08-19T16:47:52+01:00" level=info msg="Trying to refresh locks on 2 targets by 200ms" prefix="targetlocker/inmemory"
time="2020-08-19T16:47:52+01:00" level=info msg="Trying to lock 1 targets" prefix="targetlocker/inmemory"
time="2020-08-19T16:47:52+01:00" level=info msg="Trying to lock 1 targets" prefix="targetlocker/inmemory"
    TestRefreshMultiple: inmemory_test.go:154:
        	Error Trace:	inmemory_test.go:154
        	Error:      	An error is expected but got nil.
        	Test:       	TestRefreshMultiple
--- FAIL: TestRefreshMultiple (0.25s)
FAIL
```
- A variant of the above is calling `Lock` with multiple targets, if one of them is already owned by the same owner, (so it must be extended), all locks after this one are silently skipped and not locked.
- The API says `Lock` should be transactional and roll back already-locked targets when it finds an error, but it does not, it can end up partially locking things. Repro:
```
func TestLockingTransactional(t *testing.T) {
	tl := New(time.Second, time.Second)
	// lock the second target
	require.NoError(t, tl.Lock(jobID, []*target.Target{&targetTwo}))
	// try to lock both with another owner (this fails as expected)
[...]

PR #136, opened 2020-08-20
simple db based target locker

tfg13:
Summary of changes:
 - Add a minimal rdbms-based target locker implementation
 - Clarify target.Locker API docs

This adds a minimal target locker that uses a relational database as backend.

Using this locker allows sharing locks between multiple instances of Contest Server and makes locks survive server restarts.

The unit/integration test suite for this new locker is an extended version of the inmemory test suite, some tests changed to confirm with the clarified locker API and some tests added to prevent bugs like #134.

This implementation only uses target.ID as primary and only identifier for targets, we have a refactor pending to make it more clear how targets are specified and what is required to identify them.

### DIFF
--- a/plugins/targetlocker/dblocker/dblocker.go
+++ b/plugins/targetlocker/dblocker/dblocker.go
@@ -1,0 +1,343 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+//
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+package dblocker
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/facebookincubator/contest/pkg/logging"
+	"github.com/facebookincubator/contest/pkg/target"
+	"github.com/facebookincubator/contest/pkg/types"
+
+	// this blank import registers the mysql driver
+	_ "github.com/go-sql-driver/mysql"
+)
+
+// Name is the plugin name.
+var Name = "DBLocker"
+
+var log = logging.GetLogger("targetlocker/" + strings.ToLower(Name))
+
+// dblock represents parts of lock in the database, basically
+// a row from SELECT target_id, job_ID, expires_at
+type dblock struct {
+	targetID  string
+	jobID     int64
+	expiresAt time.Time
+}
+
+// String pretty-prints dblocks for logging and errors
+func (d dblock) String() string {
+	return fmt.Sprintf("target: %s job: %d expires: %s", d.targetID, d.jobID, d.expiresAt)
+}
+
+// targetIDList is a helper to convert contest targets to
+// a list of primary IDs used in the database
+func targetIDList(targets []*target.Target) []string {
+	res := make([]string, 0, len(targets))
+	for _, target := range targets {
+		res = append(res, target.ID)
+	}
+	return res
+}
+
+// listQueryString is a helper to create a (?, ?, ?) string
+// with as many ? as requested.
+// This can safely be concatenated into SQL queries as it
+// can never contain input data, it only repeates a static
+// string a given number of times.
+func listQueryString(length uint) string {
+	switch length {
+	case 0:
+		return "()"
+	case 1:
+		return "(?)"
+	default:
+		return "(" + strings.Repeat("?, ", int(length)-1) + "?)"
+	}
+}
+
+// DBLocker implements a simple target locker based on a relational database.
+// The current implementation only supports MySQL officially.
+// All functions in DBLocker are safe for concurrent use by multiple goroutines.
+type DBLocker struct {
+	driverName string
+	db         *sql.DB
+	// lockTimeout set on each initial lock request
+	lockTimeout time.Duration
+	// refreshTimeout is used during refresh
+	refreshTimeout time.Duration
+}
+
+// cleanExpired deletes all expired locks on the given targets
+// (lock owner is ignored)
+func (d *DBLocker) cleanExpired(tx *sql.Tx, targets []string, now time.Time) error {
+	q := "DELETE FROM locks WHERE expires_at < ? AND target_id IN " + listQueryString(uint(len(targets))) + ";"
+	// convert targets to a list of interface{}
+	queryList := make([]interface{}, 0, len(targets)+1)
+	queryList = append(queryList, now)
+	for _, targetID := range targets {
+		queryList = append(queryList, targetID)
+	}
+	_, err := tx.Exec(q, queryList...)
+	if err != nil {
+		return fmt.Errorf("unable to clean existing locks: %w", err)
+	}
+	return nil
+}
+
+// queryLocks returns a map of ID -> dblock for a given list of targets
+func (d *DBLocker) queryLocks(tx *sql.Tx, targets []string) (map[string]dblock, error) {
+	q := "SELECT target_id, job_id, expires_at FROM locks WHERE target_id IN " + listQueryString(uint(len(targets))) + ";"
+	// convert targets to a list of interface{}
+	queryList := make([]interface{}, 0, len(targets))
+	for _, targetID := range targets {
+		queryList = append(queryList, targetID)
+	}
+
+	rows, err := tx.Query(q, queryList...)
+	if err != nil {
+		return nil, fmt.Errorf("unable to read existing locks: %w", err)
+	}
+	defer rows.Close()
+
+	row := dblock{}
+	locks := make(map[string]dblock)
+	for rows.Next() {
+		if err := rows.Scan(&row.targetID, &row.jobID, &row.expiresAt); err != nil {
+			return nil, fmt.Errorf("unexpected read from database: %w", err)
+		}
+		locks[row.targetID] = row
+	}
+	if err := rows.Err(); err != nil {
+		return nil, fmt.Errorf("unexpected error iterating db read results: %w", err)
+	}
+	return locks, nil
+}
+
+// handleLock does the real locking, it assumes the jobID is valid
+func (d *DBLocker) handleLock(jobID int64, targets []string, timeout time.Duration) error {
+	// everything operates on this frozen time
+	now := time.Now()
+	expires := now.Add(timeout)
+	// locking is all or nothing in one transaction
+	// phantom reads might mess with the expired lock cleaning assumptions below,
+	// so request serializable isolation
+	tx, err := d.db.BeginTx(context.Background(), &sql.TxOptions{Isolation: sql.LevelSerializable})
+	if err != nil {
+		return fmt.Errorf("unable to start database transaction: %w", err)
+	}
+	defer func() {
+		// this always fails if tx.Commit() was called before, ignore error
+		_ = tx.Rollback()
+	}()
+
+	// clean expired locks first, this simplifies the logic later
+	if err := d.cleanExpired(tx, targets, now); err != nil {
+		return err
+	}
+
+	locks, err := d.queryLocks(tx, targets)
+	if err != nil {
+		return err
+	}
+	// go through existing locks, they are either held by something else (abort)
+	// held by us (update time), or not held (insert)
+	inserts := make([]string, 0)
+	updates := make([]string, 0)
+	conflicts := make([]dblock, 0)
+	for _, t := range targets {
+		lock, ok := locks[t]
+		switch {
+		case !ok:
+			inserts = append(inserts, t)
+		case lock.jobID == jobID:
+			updates = append(updates, t)
+		default:
+			conflicts = append(conflicts, lock)
+		}
+	}
+	if len(conflicts) > 0 {
+		return fmt.Errorf("unable to lock targets %v for owner %d, have conflicting locks: %v", targets, jobID, conflicts)
+	}
+
+	ins := "INSERT INTO locks (target_id, job_id, created_at, expires_at) VALUES (?, ?, ?, ?);"
+	for _, id := range inserts {
+		if _, err := tx.Exec(ins, id, jobID, now, expires); err != nil {
+			return fmt.Errorf("unable to lock target %s: %w", id, err)
+		}
+	}
+
+	upd := "UPDATE locks SET expires_at = ? WHERE target_id = ? AND job_id = ?;"
+	for _, id := range updates {
+		if _, err := tx.Exec(upd, expires, id, jobID); err != nil {
+			return fmt.Errorf("unable to refresh lock on target %s: %w", id, err)
+		}
+	}
+
+	return tx.Commit()
+}
+
+// handleUnlock does the real unlocking, it assumes the jobID is valid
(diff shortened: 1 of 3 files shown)


plugins/targetlocker/dblocker/dblocker.go
```diff
@@ -0,0 +1,310 @@
...
+	// lockTimeout set on each initial lock request
+	lockTimeout time.Duration
+	// refreshTimeout is used during refresh
+	refreshTimeout time.Duration
+}
+
+// cleanExpired deletes all expired locks on the given targets
+// (lock owner is ignored)
+func (d *DBLocker) cleanExpired(tx *sql.Tx, targets []string, now time.Time) error {
+	q := "DELETE FROM locks WHERE expires_at < ? AND target_id IN " + listQueryString(len(targets)) + ";"
```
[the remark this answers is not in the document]
insomniacslk: yeah but at least this can be modified by the caller (e.g. for integ testing), which is more future-proof

plugins/targetlocker/dblocker/dblocker.go
```diff
@@ -0,0 +1,310 @@
...
+	db         *sql.DB
+	// lockTimeout set on each initial lock request
+	lockTimeout time.Duration
+	// refreshTimeout is used during refresh
+	refreshTimeout time.Duration
+}
+
+// cleanExpired deletes all expired locks on the given targets
+// (lock owner is ignored)
+func (d *DBLocker) cleanExpired(tx *sql.Tx, targets []string, now time.Time) error {
```
insomniacslk, line 78: `len(targets)` should be checked to be > 0

plugins/targetlocker/dblocker/dblocker.go
```diff
@@ -0,0 +1,310 @@
...
+
+// listQueryString is a helper to create a (?, ?, ?) string
+// with as many ? as requested
+func listQueryString(length int) string {
+	switch length {
+	case 0:
+		return "()"
+	case 1:
+		return "(?)"
+	default:
```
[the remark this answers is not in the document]
insomniacslk: also please add a comment on why this is safe from a SQL security perspective

plugins/targetlocker/dblocker/dblocker.go
```diff
@@ -0,0 +1,310 @@
...
+		return "(" + strings.Repeat("?, ", length-1) + "?)"
+	}
+}
+
+// DBLocker implements a simple target locker based on a relational database.
+// The current implementation only supports MySQL officially.
+// All functions in DBLocker are safe for concurrent by multiple goroutines.
```
insomniacslk, line 66: typo? "for concurrent use" I assume

plugins/targetlocker/dblocker/dblocker.go
```diff
@@ -0,0 +1,310 @@
...
+	queryList := make([]interface{}, 0, len(targets))
+	for _, targetId := range targets {
+		queryList = append(queryList, targetId)
+	}
+
+	rows, err := tx.Query(q, queryList...)
+	if err != nil {
+		return nil, fmt.Errorf("unable to read existing locks: %w", err)
+	}
+	defer rows.Close()
```
insomniacslk, line 106: *cough* (just kidding, the docs say that `Close` is idempotent, and users should check `sql.Rows.Err()` instead, as documented at https://golang.org/pkg/database/sql/#Rows )

plugins/targetlocker/dblocker/dblocker.go
```diff
@@ -0,0 +1,310 @@
...
+
+	row := dblock{}
+	locks := make(map[string]dblock)
+	for rows.Next() {
+		if err := rows.Scan(&row.targetId, &row.jobID, &row.expiresAt); err != nil {
+			return nil, fmt.Errorf("unexpected read from database: %w", err)
+		}
+		locks[row.targetId] = row
+	}
```
insomniacslk, line 115: (as noted above, `rows.Err()` should be checked here)

plugins/targetlocker/dblocker/dblocker.go
```diff
@@ -0,0 +1,310 @@
...
+
+// handleLock does the real locking, it assumes the jobID is valid
+func (d *DBLocker) handleLock(jobID int64, targets []string, timeout time.Duration) error {
+	// everything operates on this frozen time
+	now := time.Now()
+	expires := now.Add(timeout)
+	// locking is all or nothing in one transaction
+	// phantom reads might mess with the expired lock cleaning assumptions below,
+	// so request serializable isolation
+	tx, err := d.db.BeginTx(context.Background(), &sql.TxOptions{Isolation: sql.LevelSerializable})
```
insomniacslk, line 127: I'm no SQL expert - does `sql.LevelSerializable` also guarantee the desired consistency level? I.e. that  lock and subsequent read will always be consistent on any primary/replica?
At a quick look isolation and consistency should not be confused, and I can't confirm that this is what we want for DB locking

plugins/targetlocker/dblocker/dblocker.go
```diff
@@ -0,0 +1,310 @@
...
+	if err := d.cleanExpired(tx, targets, now); err != nil {
+		return err
+	}
+
+	locks, err := d.queryLocks(tx, targets)
+	if err != nil {
+		return err
+	}
+	// go through existing locks, they are either held by something else (abort)
+	// help by us (update time), or not held (insert)
```
insomniacslk, line 146: "help" -> "held"

plugins/targetlocker/dblocker/dblocker.go
```diff
@@ -0,0 +1,310 @@
...
+		case !ok:
+			inserts = append(inserts, t)
+		case lock.jobID == jobID:
+			updates = append(updates, t)
+		default:
+			conflicts = append(conflicts, lock)
+		}
+	}
+	if len(conflicts) > 0 {
+		return fmt.Errorf("unable to lock, targets have conflicting locks: %v", conflicts)
```
insomniacslk, line 162: we probably also want the job ID and the target ID in this error string, for sake of debugging when necessary

insomniacslk (commented):
I have reviewed up until `handleUnlock` (excluded). Publishing this second part, then continuing

plugins/targetlocker/dblocker/dblocker.go
```diff
@@ -0,0 +1,310 @@
...
+	// detect conflicts (unlock on foreign locks) and warn about them,
+	// but don't abort.
+	conflicts := make([]dblock, 0)
+	for _, lock := range locks {
+		if lock.jobID != jobID {
+			conflicts = append(conflicts, lock)
+		}
+	}
+	if len(conflicts) > 0 {
+		log.Warningf("unable to unlock some targets due to different lock owners: %+v", conflicts)
```
insomniacslk, line 215: let's add the job ID that is failing the request here too

plugins/targetlocker/dblocker/dblocker.go
```diff
@@ -0,0 +1,310 @@
...
+	_, err := tx.Exec(q, queryList...)
+	if err != nil {
+		return fmt.Errorf("unable to clean existing locks: %w", err)
+	}
+	return nil
+}
+
+// queryLocks returns a map of ID -> dblock for a given list of targets
+func (d *DBLocker) queryLocks(tx *sql.Tx, targets []string) (map[string]dblock, error) {
+	q := "SELECT target_id, job_id, expires_at FROM locks WHERE target_id IN " + listQueryString(len(targets))
```
insomniacslk, line 95: the trailing ";" is missing here and present elsewhere, and we should probably be consistent

plugins/targetlocker/dblocker/dblocker.go
```diff
@@ -0,0 +1,310 @@
...
+	// only remove locks held by the owner
+	del := "DELETE FROM locks WHERE job_id = ? AND target_id IN " + listQueryString(len(targets)) + ";"
+	queryList := make([]interface{}, 0, len(targets)+1)
+	queryList = append(queryList, jobID)
+	for _, targetId := range targets {
+		queryList = append(queryList, targetId)
+	}
+	_, err = tx.Exec(del, queryList...)
+	if err != nil {
+		return fmt.Errorf("unable to unlock targets: %w", err)
```
insomniacslk, line 227: and job ID here too

plugins/targetlocker/dblocker/dblocker.go
```diff
@@ -0,0 +1,310 @@
...
+	}
+	log.Debugf("Requested to refresh %d targets for job ID %d: %v", len(targets), jobID, targets)
+
+	return d.handleLock(int64(jobID), targetIDList(targets), d.refreshTimeout)
+}
+
+// ResetAllLocks resets the database and clears all locks, regardless of who owns them.
+// This is primarily for testing, and should not be used by used in prod, this
+// is why it is not exposed by target.Locker
+func (d *DBLocker) ResetAllLocks() error {
```
insomniacslk, line 269: being internal, unexport it?

plugins/targetlocker/dblocker/dblocker.go
tfg13: Oh good point!

plugins/targetlocker/dblocker/dblocker.go
tfg13: Not sure I am following here. I am using this purely for query isolation, since some steps assume previous steps are done properly. Example: When locking I read other locks and check them (because we need this to produce a useful error anyway). The DB will enforce only one lock per asset on its own of course, but it makes the code simple to assume there won't be any conflicts after I run the check.

What are you asking about exactly wrt. to consistency, do you mean strict read-after-write consistency for multiple queries to the same DB?

plugins/targetlocker/dblocker/dblocker.go
tfg13: They are! This is why there is a pretty printer on the dblock struct :)

plugins/targetlocker/dblocker/dblocker.go
tfg13: same as above

plugins/targetlocker/dblocker/dblocker.go
tfg13: No then the semi-external integration tests can't reach it :/

plugins/targetlocker/dblocker/dblocker.go
tfg13: I just checked the official examples - it is also inconsistent there. AAAAA
Fixing :)

insomniacslk: yeah :( it is safe with mysql and probably other engines too, but I guess it doesn't hurt to avoid bad karma :)

plugins/targetlocker/dblocker/dblocker.go
insomniacslk: we discussed this internally already, and concluded that transactions paired to serializable isolation should be good enough for locking purposes. To answer to your question, yes, I meant strict read-after-write consistency

plugins/targetlocker/dblocker/dblocker.go
insomniacslk: I mean not just the info on what is currently holding the lock, but also on what tried to get the lock but failed

plugins/targetlocker/dblocker/dblocker.go
insomniacslk: same as above :P

plugins/targetlocker/dblocker/dblocker.go
tfg13: done, looks like this now:
```
unable to lock targets [001 002] for owner 124, have conflicting locks: [target: 001 job: 123 expires: 2020-08-25 15:01:48 +0000 UTC target: 002 job: 123 expires: 2020-08-25 15:01:48 +0000 UTC]
```

plugins/targetlocker/dblocker/dblocker.go
tfg13: done, same as above :P

plugins/targetlocker/dblocker/dblocker.go
```diff
@@ -0,0 +1,310 @@
...
+func (d dblock) String() string {
+	return fmt.Sprintf("target: %s, job %d, expires %s", d.targetId, d.jobID, d.expiresAt)
+}
+
+// targetIDList is a helper to convert contest targets to
+// a list of primary IDs used in the database
+func targetIDList(targets []*target.Target) []string {
+	res := make([]string, 0, len(targets))
+	for _, target := range targets {
+		res = append(res, target.ID)
```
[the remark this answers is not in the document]
tfg13: ok I am bringing the verify function back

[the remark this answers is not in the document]
tfg13: added `validateTargets` to make sure none are empty

plugins/targetlocker/dblocker/dblocker.go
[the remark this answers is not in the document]
tfg13: I am going to punt this one to a future PR. It opens the "concat query strings" can of worms again, and at that point we probably should use something like safesql, but that is too big a refactor to integrate it now.

plugins/targetlocker/dblocker/dblocker.go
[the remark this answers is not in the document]
tfg13: also added a unit test

tfg13:
Ok addressed all comments except the one about making the table name a parameter, punting this one as we probably should use safesql then, which is a bigger rework.

plugins/targetlocker/dblocker/dblocker.go
```diff
@@ -0,0 +1,310 @@
...
+// dblock represents parts of lock in the database, basically
+// a row from SELECT target_id, job_ID, expires_at
+type dblock struct {
+	targetId  string
+	jobID     int64
+	expiresAt time.Time
+}
+
+// String pretty-prints dblocks for logging and errors
+func (d dblock) String() string {
```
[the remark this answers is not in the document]
insomniacslk: I prefer the latter :P (also, that's `%+v`, not `%#+v`) but it's personal, I'm OK with the current one

plugins/targetlocker/dblocker/dblocker.go
insomniacslk: this is marked as resolved - but why don't you check for zero length, instead of executing the DB query anyway?
```
if len(targets) == 0 {
    return nil
}
```

insomniacslk (approved):

plugins/targetlocker/dblocker/dblocker.go
tfg13: I do, but right away in `Unlock`, `Lock`, and `RefreshLocks`, so we won't get here without locks :)

Reason I put it there is that there are 2 other places where it can create invalid queries with `... IN ()` (basically everywhere `listQueryString` is used)

PR #138, opened 2020-08-21
[test_runner] Fix deadlock in routing code

marcoguerri:
Routing now consists of input routing and output routing.
If any of the two blocks fails, as we are seeing now with a bug
in ForEachTarget which serializes targets injection, the error should
be considered as fatal, and a termination signal should be propagated
to all the routing logic, the router should return and trigger a
cancellation of the whole test.

Signed-off-by: Marco Guerri <marco.guerri.dev@fastmail.com>


marcoguerri:
So, this change ended up pulling in several `go fmt` changes that are not related. The relevant change is limited to `route` in `test_runner_route.go`. Let me know if you want me to split into two separate changes.

Upstream test is done via usual integration tests. For internal tests (where we are actually seeing the deadlock due to long runtime of one plugin), see gist number 139572125.

pkg/runner/test_runner_route.go
```diff
@@ -171,125 +180,150 @@ func (r *stepRouter) emitOutEvent(t *target.Target, err error) error {
...
+	if err != nil {
+		log.Debugf("routeOut failed: %w", err)
+		return 0, err
+	}
+	return len(egressTarget), nil
 
+}
 
+// route implements the routing logic from the previous routing block to the test step
+// and from the test step to the next routing block
```
tfg13, line 257: duplicate comment :)

pkg/runner/test_runner_route.go
```diff
@@ -171,125 +180,148 @@ func (r *stepRouter) emitOutEvent(t *target.Target, err error) error {
...
+		routingErr                          error
+		routeInCompleted, routeOutCompleted bool
+	)
+	terminateInternal := make(chan struct{})
+	errRouteInCh := make(chan error)
+	errRouteOutCh := make(chan error)
+
+	go func() {
+		var err error
+		inTargets, err = r.routeIn(terminateInternal)
```
hugelgupf, line 271: Why is there a terminate internal? You could just pass terminate through to every one of these.

marcoguerri: My intention is for `terminate` to be owned by the caller, which shouldn't have to bother to check if has been closed already. So only the owner can close it.

marcoguerri: Our rationale behind moving away from context is more or less aligned with https://dave.cheney.net/2017/08/20/context-isnt-for-cancellation. We were planning to move to https://github.com/insomniacslk/interruption, but never happened.

PR #139, opened 2020-08-21
Add tests for pkg/teststeps.ForEachTarget, fix non-parallel execution

insomniacslk:
@dimkup has found that ForEachTarget is not parallel as it is supposed
to be. I am adding tests for this function, and fixing the bug that
makes this function not behave as documented.

### DIFF
--- a/plugins/teststeps/teststeps.go
+++ b/plugins/teststeps/teststeps.go
@@ -18,7 +18,7 @@ var log = logging.GetLogger("plugins/teststeps")
 // ForEachTarget function below.
 type PerTargetFunc func(cancel, pause <-chan struct{}, target *target.Target) error
 
-// ForEachTarget is a facility provided to simplify plugin implemtations. This
+// ForEachTarget is a facility provided to simplify plugin implementations. This
 // function wraps the logic that handles target routing through the in/out/err
 // test step channels, and also handles cancellation and pausing.
 // The user of this function has to pass the cancel, pause, and test step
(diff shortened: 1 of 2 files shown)


plugins/teststeps/teststeps.go
```diff
@@ -27,49 +27,46 @@ type PerTargetFunc func(cancel, pause <-chan struct{}, target *target.Target) er
...
+				case <-pause:
+					log.Debugf("%s: ForEachTarget: received pausing signal", pluginName)
+					return nil
 				}
-			case <-cancel:
-				return nil
-			case <-pause:
-				return nil
 			}
 		case <-cancel:
```
dimkup, line 71: Do we want to handle cancel/pause signals, while f() is running? It's necessary while we are trying to write target/error to the out channels, but here I guess plugin implementation should care about cancel/pauses signals (otherwise, why do we pass them to f() as a parameters?). Also it introduces uncertainty in behaviour: depending on  which goroutine (ForEachTarget or f()) got the signal, it could be handled in a different way, which is matter of chance.

plugins/teststeps/teststeps.go
```diff
@@ -27,49 +27,46 @@ type PerTargetFunc func(cancel, pause <-chan struct{}, target *target.Target) er
...
+	type tgtErr struct {
+		target *target.Target
+		err error
+	}
+	tgtErrCh := make(chan tgtErr)
 	for {
 		select {
-		case target := <-ch.In:
-			if target == nil {
+		case tgt := <-ch.In:
```
dimkup, line 37: Could it be that ForEachTarget has read nil from the In channel _before_ all the targets processed? It would lead to leaking the targets..

plugins/teststeps/teststeps_test.go
```diff
@@ -0,0 +1,254 @@
...
+		maxWaitTime := sleepTime * 3
+		deadline := time.Now().Add(maxWaitTime)
+		log.Printf("Setting deadline to now+%s", maxWaitTime)
+		for {
+			select {
+			case <-d.done:
+				break
+			case tgt := <-d.outCh:
+				t.Errorf("Step for target %+v expected to fail but completed successfully instead", tgt)
+			case err := <-d.errCh:
```
dimkup, line 238: Probably it would be good to count targets here, just to be sure that _all_ of them have been processed.

PR #140, opened 2020-08-24
Auto-select target locker

tfg13:
Keeping this in preview mode, just so that we don't forget. To be merged once we have some confidence in the new locker, and the target rework is done.


tfg13:
Also this will need to be rebased because github doesn't understand stacks, or at least I don't know where they are

tfg13:
Still blocked. The DB locker is used and works, but does more DB queries so takes a little bit longer. We need the change to support per-job targetmanager timeouts first. @fabriziocucci wanted to look into this IIRC

tfg13:
Ok, reworked this. It will now follow the main DB selection by default, but it still can be set manually

mimir-d (approved):

PR #142, opened 2020-09-03
Added transaction handling to inmemory locks

Richard-Oak:
Fixes issue #134

Add a parallel map of locks containing the state of the locks
after the transaction completes successfully. In the event of
an error, this map is discarded. Otherwise, it is merged into
the main lock map at the end.


tfg13 (approved):
LGTM

PR #143, opened 2020-09-07
For each target tests

dimkup:
This is a rework of https://github.com/facebookincubator/contest/pull/139 by @insomniacslk. Current implementation of ForEachTarget helper handles incoming targets sequentially, but the task runner expects them to be handled in parallel and has a timeout for each target to be consumed (30 sec). This PR makes  ForEachTarget handle targets in parallel, as it should be by design.

### DIFF
--- a/plugins/teststeps/teststeps.go
+++ b/plugins/teststeps/teststeps.go
@@ -42,23 +42,32 @@ func ForEachTarget(pluginName string, cancel, pause <-chan struct{}, ch test.Tes
 			log.Debugf("%s: ForEachTarget: exiting incoming loop, targets inflight %d", pluginName, tgtInFlight)
 		}()
 		for {
-			tgt := <-ch.In
-			log.Debugf("%s: ForEachTarget: received target %s", pluginName, tgt)
+			select {
+			case tgt := <-ch.In:
+				log.Debugf("%s: ForEachTarget: received target %s", pluginName, tgt)
+				if tgt == nil {
+					log.Debugf("%s: ForEachTarget: all targets have been received", pluginName)
+					noMoreTargetsCh <- struct{}{}
+					return
+				}
 
-			if tgt == nil {
-				log.Debugf("%s: ForEachTarget: all targets have been received", pluginName)
-				noMoreTargetsCh <- struct{}{}
+				go func() {
+					tgtErrCh <- tgtErr{target: tgt, err: f(cancel, pause, tgt)}
+				}()
+				atomic.AddInt32(tgtInFlight, 1)
+			case <-cancel:
+				log.Debugf("%s: ForEachTarget: incoming loop canceled", pluginName)
+				return
+			case <-pause:
+				log.Debugf("%s: ForEachTarget: incoming loop paused", pluginName)
 				return
 			}
 
-			go func() {
-				tgtErrCh <- tgtErr{target: tgt, err: f(cancel, pause, tgt)}
-			}()
-			atomic.AddInt32(tgtInFlight, 1)
 		}
 	}(&tgtInFlight)
 
 	noMoreTargets := false
+	reportResults := true
 	defer func() {
 		log.Debugf("%s: ForEachTarget: exiting outgoing loop, targets inflight %d, the last target received %v", pluginName, tgtInFlight, noMoreTargets)
 	}()
@@ -66,30 +75,34 @@ func ForEachTarget(pluginName string, cancel, pause <-chan struct{}, ch test.Tes
 		select {
 		case te := <-tgtErrCh:
 			atomic.AddInt32(&tgtInFlight, -1)
-			if te.err != nil {
-				select {
-				case ch.Err <- cerrors.TargetError{Target: te.target, Err: te.err}:
-					log.Errorf("%s: ForEachTarget: failed to apply test step function on target %s: %v", pluginName, te.target, te.err)
-				case <-cancel:
-					log.Debugf("%s: ForEachTarget: received cancellation signal", pluginName)
-					return nil
-				case <-pause:
-					log.Debugf("%s: ForEachTarget: received pausing signal", pluginName)
-					return nil
+			if reportResults {
+				if te.err != nil {
+					select {
+					case ch.Err <- cerrors.TargetError{Target: te.target, Err: te.err}:
+						log.Errorf("%s: ForEachTarget: failed to apply test step function on target %s: %v", pluginName, te.target, te.err)
+					case <-cancel:
+						log.Debugf("%s: ForEachTarget: received cancellation signal while reporting error", pluginName)
+						reportResults = false
+					case <-pause:
+						log.Debugf("%s: ForEachTarget: received pausing signal while reporting error", pluginName)
+						reportResults = false
+					}
+				} else {
+					select {
+					case ch.Out <- te.target:
+						log.Debugf("%s: ForEachTarget: target %s completed successfully", pluginName, te.target)
+					case <-cancel:
+						log.Debugf("%s: ForEachTarget: received cancellation signal while reporting success", pluginName)
+						reportResults = false
+					case <-pause:
+						log.Debugf("%s: ForEachTarget: received pausing signal while reporting success", pluginName)
+						reportResults = false
+					}
 				}
 			} else {
-				select {
-				case ch.Out <- te.target:
-					log.Debugf("%s: ForEachTarget: target %s completed successfully", pluginName, te.target)
-				case <-cancel:
-					log.Debugf("%s: ForEachTarget: received cancellation signal", pluginName)
-					return nil
-				case <-pause:
-					log.Debugf("%s: ForEachTarget: received pausing signal", pluginName)
-					return nil
-				}
+				log.Debugf("%s: ForEachTarget: the result is ignored due to cancelation: %v", pluginName, te)
 			}
-			if atomic.LoadInt32(&tgtInFlight) == 0 && noMoreTargets {
+			if atomic.LoadInt32(&tgtInFlight) == 0 && (noMoreTargets || !reportResults) {
 				return nil
 			}
 		case <-noMoreTargetsCh:
@@ -99,11 +112,11 @@ func ForEachTarget(pluginName string, cancel, pause <-chan struct{}, ch test.Tes
 			}
 			noMoreTargets = true
 		case <-cancel:
-			log.Debugf("%s: ForEachTarget: received cancellation signal", pluginName)
-			return nil
+			log.Debugf("%s: ForEachTarget: received cancellation signal while waiting for results", pluginName)
+			reportResults = false
 		case <-pause:
-			log.Debugf("%s: ForEachTarget: received pausing signal", pluginName)
-			return nil
+			log.Debugf("%s: ForEachTarget: received pausing signal while waiting for results", pluginName)
+			reportResults = false
 		}
 	}
 }
--- a/plugins/teststeps/teststeps.go
+++ b/plugins/teststeps/teststeps.go
@@ -6,6 +6,8 @@
 package teststeps
 
 import (
+	"sync/atomic"
+
 	"github.com/facebookincubator/contest/pkg/cerrors"
 	"github.com/facebookincubator/contest/pkg/logging"
 	"github.com/facebookincubator/contest/pkg/target"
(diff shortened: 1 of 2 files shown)


plugins/teststeps/teststeps.go
```diff
@@ -27,53 +29,80 @@ type PerTargetFunc func(cancel, pause <-chan struct{}, target *target.Target) er
...
+	}()
+	for {
+		select {
+		case te := <-tgtErrCh:
+			atomic.AddInt32(&tgtInFlight, -1)
+			if te.err != nil {
+				select {
+				case ch.Err <- cerrors.TargetError{Target: te.target, Err: te.err}:
+					log.Errorf("%s: ForEachTarget: failed to apply test step function on target %s: %v", pluginName, te.target, te.err)
+				case <-cancel:
```
marcoguerri, line 73: I think this will be problematic. One of the requirements from the plugins perspective is that once they receive cancel/pause, they should begin returning immediately and if we don't wait for the various `f` to actually return, we'll be hiding plugins that actually do not honor cancellation signal. The runner higher up will set a timeout for this to happen, and fail if the plugins haven't returned.

plugins/teststeps/teststeps.go
```diff
@@ -27,53 +29,80 @@ type PerTargetFunc func(cancel, pause <-chan struct{}, target *target.Target) er
...
+	defer func() {
+		log.Debugf("%s: ForEachTarget: exiting outgoing loop, targets inflight %d, the last target received %v", pluginName, tgtInFlight, noMoreTargets)
+	}()
+	for {
+		select {
+		case te := <-tgtErrCh:
+			atomic.AddInt32(&tgtInFlight, -1)
+			if te.err != nil {
+				select {
+				case ch.Err <- cerrors.TargetError{Target: te.target, Err: te.err}:
```
marcoguerri, line 71: I am hoping to simplify this interface with https://github.com/facebookincubator/contest/pull/121, so that there will be only a unified channel, where an error might be nil. Having two channels requires always having an else branch, which makes things significantly more difficult to read and to maintain.

plugins/teststeps/teststeps.go
marcoguerri: So, I think here we should have also a wait group?

plugins/teststeps/teststeps.go
```diff
@@ -27,53 +29,80 @@ type PerTargetFunc func(cancel, pause <-chan struct{}, target *target.Target) er
...
-						return nil
-					}
+	noMoreTargets := false
+	defer func() {
+		log.Debugf("%s: ForEachTarget: exiting outgoing loop, targets inflight %d, the last target received %v", pluginName, tgtInFlight, noMoreTargets)
+	}()
+	for {
+		select {
+		case te := <-tgtErrCh:
+			atomic.AddInt32(&tgtInFlight, -1)
```
marcoguerri, line 68: Do we really need to keep track of `tgtInFlight`? Since I think we need a wait group, could we just wait for `ch.In` to close, and then wait for the wait group to terminate?

plugins/teststeps/teststeps.go
dimkup: I didn't know that we expect all the ```f``` instances should return something. Was thinking that the cancel signal should be handled in "fire and forget" fashion. Will fix.

dimkup: Do I understand right that a plugin should return errors for all "inflight" targets in case of cancelation?

plugins/teststeps/teststeps.go
dimkup: It would be nice to use simplified channels here, but does it mean PR should wait for #121 to be accepted?

plugins/teststeps/teststeps.go
dimkup: Yeah, we could use a wait group here in this case.

plugins/teststeps/teststeps.go
marcoguerri: Hehe, much worse #121 to be accepted, which depends on #119, which depends #118, which depends on having a tool to do schema migrations. Lots of work to unblock #121 still to do, and I don't foresee to be able to get it done in the short term. This should definitely go out before #121, it was mostly an FYI.

plugins/teststeps/teststeps.go
marcoguerri: So we don't expect `f` to return anything in particular, this at the moment is not really specified, but we do expect the main `Run` function of the plugin to return, and to do so as a signal that the internal plugin logic has been interrupted.

The runner at the moment ignores any error returning from the plugin after cancellation, but it does flag a plugin as non-returning if it doesn't return within a certain timeout after the cancellation signal. 

Does it make sense?

dimkup: So we just want to detect plugins which do not respect the cancelation signal? Ok, will fix.

dimkup: Didn't manage to use a work group there, because there is no way to check remaining value in a work group without blocking. wg.Wait() would block the goroutine which consumes results from ```f``` instances and they would not be finished ever. Of course we cloud start another "outgoing" goroutine which consumes and reports results from ```f``` but I think it would make things even more complex.. Pleas let me know if you would like me to try this approach.

plugins/teststeps/teststeps.go
```diff
@@ -27,54 +29,94 @@ type PerTargetFunc func(cancel, pause <-chan struct{}, target *target.Target) er
...
 					case <-pause:
-						log.Debugf("%s: ForEachTarget: received pausing signal", pluginName)
-						return nil
+						log.Debugf("%s: ForEachTarget: received pausing signal while reporting success", pluginName)
+						reportResults = false
 					}
 				}
-			case <-cancel:
+			} else {
+				log.Debugf("%s: ForEachTarget: the result is ignored due to cancelation: %v", pluginName, te)
```
xaionaro, line 103: nitpicking:

A typo, should be `cancellation`.

xaionaro (changes requested):
Looks good. Please address the comment marked as "error". And I haven't check unit-tests, sorry :)

xaionaro:
Also please rebase this PR on top of Marco's changes to pipelines.

plugins/teststeps/teststeps_test.go
```diff
@@ -0,0 +1,369 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
```
xaionaro, line 1: I see these tests takes long:
```
xaionaro@void:~/go/src/github.com/facebookincubator/contest$ go test ./plugins/teststeps/ -count=1
ok      github.com/facebookincubator/contest/plugins/teststeps  4.343s
```

May be we should add a build tag like `longtests` or something like that. @marcoguerri : what do you think?

plugins/teststeps/teststeps.go
```diff
@@ -27,54 +29,94 @@ type PerTargetFunc func(cancel, pause <-chan struct{}, target *target.Target) er
...
+		select {
+		case te := <-tgtErrCh:
+			atomic.AddInt32(&tgtInFlight, -1)
+			if reportResults {
+				if te.err != nil {
 					select {
-					case ch.Err <- cerrors.TargetError{Target: target, Err: err}:
-						log.Errorf("%s: ForEachTarget: failed to apply test step function on target %s: %v", pluginName, target, err)
+					case ch.Err <- cerrors.TargetError{Target: te.target, Err: te.err}:
+						log.Errorf("%s: ForEachTarget: failed to apply test step function on target %s: %v", pluginName, te.target, te.err)
```
xaionaro, line 82: **warning:**

This `Errorf` should be before `select`, if I understand correctly.

plugins/teststeps/teststeps.go
```diff
@@ -27,54 +29,94 @@ type PerTargetFunc func(cancel, pause <-chan struct{}, target *target.Target) er
...
-						return nil
+						log.Debugf("%s: ForEachTarget: received pausing signal while reporting error", pluginName)
+						reportResults = false
 					}
 				} else {
 					select {
-					case ch.Out <- target:
-						log.Debugf("%s: ForEachTarget: target %s completed successfully", pluginName, target)
+					case ch.Out <- te.target:
+						log.Debugf("%s: ForEachTarget: target %s completed successfully", pluginName, te.target)
```
xaionaro, line 93: **warning:**

This `Debugf` should be before `select`, if I understand correctly.

plugins/teststeps/teststeps.go
```diff
@@ -27,54 +29,94 @@ type PerTargetFunc func(cancel, pause <-chan struct{}, target *target.Target) er
...
+				log.Debugf("%s: ForEachTarget: received target %s", pluginName, tgt)
+				if tgt == nil {
+					log.Debugf("%s: ForEachTarget: all targets have been received", pluginName)
+					noMoreTargetsCh <- struct{}{}
+					return
+				}
+
+				go func() {
+					tgtErrCh <- tgtErr{target: tgt, err: f(cancel, pause, tgt)}
+				}()
```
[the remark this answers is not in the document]
marcoguerri: Ok I see. I think in the solution we were discussed in https://play.golang.org/p/Kvki6Z2RNHm, we were serializing read-write, i.e.

```
readValue <- chR
chW <- readValue
```

But I think this would work, just spawn a  background goroutine for each incoming target. Yeah, not sure why I haven't gone with this solution before, it seems so trivial now (I have everything in place, like workground, goroutine, etc. except that I serialize the goroutine...) :( 
I will give it a try, thank you.

plugins/teststeps/teststeps_test.go
```diff
@@ -0,0 +1,369 @@
...
+			d.inCh <- &target.Target{Name: fmt.Sprintf("target%00d", i)}
+		}
+		// signal end of input
+		d.inCh <- nil
+	}()
+	go func() {
+		for {
+			select {
+			case <-d.done:
+				break
```
xaionaro, line 123: It seems to be you meant `return` here (and in other tests).

plugins/teststeps/teststeps_test.go
```diff
@@ -0,0 +1,369 @@
...
+	sleepTime := time.Second * 5
+	numTargets := 10
+	var canceledTargets int32
+	d := newData()
+
+	fn := func(cancel, pause <-chan struct{}, tgt *target.Target) error {
+		log.Printf("Handling target %+v", tgt)
+		select {
+		case <-cancel:
+			log.Printf("target %+v caneled", tgt)
```
xaionaro, line 340: **nitpicking:**

A typo, should be: `cancelled`.

plugins/teststeps/teststeps.go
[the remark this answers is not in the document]
xaionaro: > I will give it a try, thank you.

I meant only to humbly ask to try to find a simpler solution for future code, I haven't meant to spent time on redoing what is already achieved :)

plugins/teststeps/teststeps.go
```diff
@@ -27,54 +29,94 @@ type PerTargetFunc func(cancel, pause <-chan struct{}, target *target.Target) er
...
 			}
+			noMoreTargets = true
 		case <-cancel:
-			return nil
+			log.Debugf("%s: ForEachTarget: received cancellation signal while waiting for results", pluginName)
+			reportResults = false
 		case <-pause:
-			return nil
+			log.Debugf("%s: ForEachTarget: received pausing signal while waiting for results", pluginName)
+			reportResults = false
```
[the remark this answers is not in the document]
marcoguerri: I tend to agree that the second for loop should be collapsed and it would simplify a lot. Also, if we agree to use a `wg` for waiting that all goroutines terminate, then the final wait should not even need to select on cancellation signals. We expect `f` to return on cancellation, if they don't, there a bug and we are leaking stuff behind anywa.

plugins/teststeps/teststeps.go
```diff
@@ -27,54 +29,94 @@ type PerTargetFunc func(cancel, pause <-chan struct{}, target *target.Target) er
...
+			case tgt := <-ch.In:
+				log.Debugf("%s: ForEachTarget: received target %s", pluginName, tgt)
+				if tgt == nil {
+					log.Debugf("%s: ForEachTarget: all targets have been received", pluginName)
+					noMoreTargetsCh <- struct{}{}
+					return
+				}
+
+				go func() {
+					tgtErrCh <- tgtErr{target: tgt, err: f(cancel, pause, tgt)}
```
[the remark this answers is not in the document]
xaionaro: Yeah, I think you right, my fault, sorry.

plugins/teststeps/teststeps.go
dimkup: I was thinking that we should log the result only after it was consumed by the upper layer, but probably you are right, will fix.

plugins/teststeps/teststeps_test.go
dimkup: Most of the tests were created by @insomniacslk, but I'll check :)

plugins/teststeps/teststeps.go
xaionaro: May me I wrong. OK, here's another approach:
* We should move it upper.
* OR we should change the text to signal that the error was sent.

Just the problem is that the current text is about something should be signaled before `select`, IMHO.

plugins/teststeps/teststeps.go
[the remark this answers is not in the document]
xaionaro: > I didn't have a goal to rewrite the the helper completely, but just to add some properties we need.

:+1: 

Let's leave as is. But in next iteration it would be nice to make an attempt to simplify this.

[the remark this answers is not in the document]
xaionaro: > Of course we could send the results directly, but Run method has to return only after all the processing routines are finished, so we need to track them anyway. Also in this case we would need to handle cancel/pause logic inside of processing routines. So it's definitely doable, but would not be much concise any way, imho. :)

I made an attempt to prepare a proof of concept (without `for`):
https://dpaste.org/FkAC

And I think it could be further simplified and improved. Never mind, it was just a thought out loud. Let's leave as is :)

xaionaro (approved):

xaionaro:
@dimkup : Accepting in advance, but please rebase the PR on top of Marco's changes :)

xaionaro:
@dimkup : After discussion with @marcoguerri I realized it is too long to wait. Let's just merge as is. Sorry for this mess :)

plugins/teststeps/teststeps.go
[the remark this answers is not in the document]
dimkup: Heh, I was wrong. :) Your approach is 2x times shorted and it looks like it satisfies all the requirements. So we probably should go for it. Let it be my tech debt.. :)

PR #144, opened 2020-09-16
Fix race condition in job status emission

insomniacslk:
Before this patch there is a race condition where the job status event
is emitted before a report is saved. A client waiting on the status
might get zero job reports if the job status is fetched before the
report is saved to the database. This is a behaviour we have seen
internally.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ global:
 env:
   global:
     - COMPOSE_VERSION=1.25.4
+    - GO111MODULE=on
   jobs:
     - TEST_SUITE=tests
     - TEST_SUITE=linters
--- a/pkg/jobmanager/start.go
+++ b/pkg/jobmanager/start.go
@@ -69,6 +69,19 @@ func (jm *JobManager) start(ev *api.Event) *api.EventResponse {
 			return
 		}
 
+		// store job report before emitting the job status event, to avoid a
+		// race condition when waiting on a job status where the event is marked
+		// as completed but no report exists.
+		jobReport := job.JobReport{
+			JobID:        j.ID,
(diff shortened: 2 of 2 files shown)


insomniacslk:
Errored for an unrelated issue:
```
go get: warning: modules disabled by GO111MODULE=auto in GOPATH/src;
	ignoring go.mod;
	see 'go help modules'
github.com/insomniacslk/termhook
# github.com/insomniacslk/termhook
../../insomniacslk/termhook/termhook.go:98:55: cannot use &a (type *syscall.Termios) as type *unix.Termios in argument to termios.Tcgetattr
../../insomniacslk/termhook/termhook.go:101:38: cannot use (*syscall.Termios)(&a) (type *syscall.Termios) as type *unix.Termios in argument to termios.Cfmakeraw
../../insomniacslk/termhook/termhook.go:102:71: cannot use &a (type *syscall.Termios) as type *unix.Termios in argument to termios.Tcsetattr
```

tfg13 (approved):

insomniacslk:
The new error is
```
--- FAIL: TestJobManagerSuiteMemoryStorage (16.05s)
    --- FAIL: TestJobManagerSuiteMemoryStorage/TestJobManagerJobCrash (5.01s)
        common.go:405: 
            	Error Trace:	common.go:405
            	Error:      	Not equal: 
            	            	expected: 1
            	            	actual  : 0
            	Test:       	TestJobManagerSuiteMemoryStorage/TestJobManagerJobCrash
FAIL
```

With some debugging output it shows that `pollForEvent` returns zero events after all the attempts have exhausted (and in other words, that `eventManager.Fetch` returns no event of type `EventJobFailed`)

PR #145, opened 2020-09-21
Basic parameters length checks

dimkup:
We need to validate some input data in order to be sure that they fits underlaying storage structures. Otherwise we could get run time errors or unpredictable behaviour. This PR adds basic length checks for some fields like JobName, TestName, TestStepLabel etc..


.travis.yml
```diff
@@ -1,5 +1,8 @@
 language: go
 
+go:
+  - "1.13"
```
insomniacslk, line 4: I think we can safely jump to 1.15

pkg/api/api.go
```diff
@@ -51,14 +52,18 @@ func New(serverIDFunc func() string) *API {
...
+	} else {
+		if hn, err := os.Hostname(); err == nil {
+			serverID = hn
+		}
 	}
-	hn, err := os.Hostname()
-	if err != nil {
-		return "<unknown>"
+	if err := limits.Validator.ValidateServerID(serverID); err != nil {
+		panic(err)
```
insomniacslk, line 64: definitely not a fan of panicking in library code - how complex is it to change `ServerID`'s signature instead?

pkg/storage/limits/limits.go
```diff
@@ -0,0 +1,107 @@
...
+	MaxLen    int
+	ActualLen int
+}
+
+func (err ErrParameterIsTooLong) Error() string {
+	return fmt.Sprintf("%s is too long: %d > %d", err.DataName, err.ActualLen, err.MaxLen)
+}
+
+// Validator is an instance of current storage limits validator
+var Validator LimitsValidator = &BasicStorageLimitsValidator{}
```
insomniacslk, line 41: I would steer away from global variables - can we instead have a `NewValidator` function that returns `(LimitsValidator, error)` instead? Testing is equally easier, and we move away from global state

pkg/storage/limits/limits.go
```diff
@@ -0,0 +1,107 @@
...
+
+// BasicStorageLimitsValidator is a simple LimitsValidator implementation
+type BasicStorageLimitsValidator struct{}
```
insomniacslk, line 44: what makes it "basic" ? Do we expect other implementations other than for testing?

.travis.yml
dimkup: We have "You will also need a recent version of Go (we recommend at least Go 1.13 at the moment)." in Readme, so I thought that 1.13 is a minimal supported version. If so, I think it's wise to keep tests running the minimal version. :)

pkg/storage/limits/limits.go
dimkup: In fact Validator does not have any state.. I just don't like to create instances for each call. We also would need to handle an error every time. I don't see any reason NewValidator could fail, so could it be func NewValidator() LimitsValidator? 
Regarding tests, it would be fairly easy to put mocked validator in to the global var, but I don't see any way to mock value returned by NewValidator, if its called as limits.NewValidator().Validate..(). May be we could use factory here as a tradeoff?

pkg/storage/limits/limits.go
dimkup: I just wanted some name for the interface implementation. I was thinking to call the interface ILimitsValidator or LimitsValidatorInterface, but decided that's better to add something to the implementation name and didn't come up with anything better than "basic":))

pkg/api/api.go
dimkup: I agree, but it's not clear, where ServerID() can be called. It looks like "general getter". Imho it would be better to calculate the serverId explicitly in some initialisation code, validate it and then just return to every call. In order to do it immutable, we can warp in in lambda. What do you think?

pkg/storage/limits/limits.go
xaionaro: Out of curiosity, why do we need an interface? Just in case: https://github.com/golang/go/wiki/CodeReviewComments#interfaces

xaionaro (changes requested):
Sorry for this nitpicking, but requesting changes due to `panic` :)

.travis.yml
insomniacslk: We can also bump the one in the README :) I'm definitely not willing to support Go < 1.15 at the moment

pkg/storage/limits/limits.go
insomniacslk: a validator as global variable is essentially state, I can use a different term if "state" doesn't feel right. Anyway, subsequent calls depend on the value pointed to by `Validator`, which can be changed, so that such calls may end up using a different object. Testing is an example where this becomes challenging: you cannot run tests in parallel if you want to use different Validator implementations, because you don't have a guarantee that another test hasn't changed it while you're using it

pkg/storage/limits/limits.go
insomniacslk: Would testing be impacted if we move away from an interface?

pkg/api/api.go
```diff
@@ -51,14 +52,18 @@ func New(serverIDFunc func() string) *API {
...
 // generation function can be passed to New().
 func (a API) ServerID() string {
+	serverID := "<unknown>"
 	if a.serverIDFunc != nil {
-		return a.serverIDFunc()
+		serverID = a.serverIDFunc()
+	} else {
+		if hn, err := os.Hostname(); err == nil {
+			serverID = hn
+		}
```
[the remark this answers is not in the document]
insomniacslk: This should be OK: `serverID` is returned regardless of this code. If `err !=` this code will simply skip to the next block (line 63, where `ValidateServerID` is called) and will eventually return `serverID` containing "<unknown>" from line 55

pkg/storage/limits/limits.go
dimkup: I understand problems of shared state, but I don't see a good way to solve them, besides using context propagation, as @xaionaro was pointing out. Factory does not necessarily return a new instance of an object. It returns _some_ instance and caller should not care if it's a new one or not. If I see New... function I expect to get a new instance from it. If I see Get... function I know that I cannot be sure. My understanding of factory is that it's a manageable source of instances, which allows to change instances creation strategy easily. So it would be GetValidator instead of NewValidator. :)

pkg/storage/limits/limits.go
dimkup: I'm not familiar with interface usage patterns in go, sorry:) I was thinking that by using interface we could easily implement multiple validators, or say make storage plugin to return its own implementation, thus make the limits storage plugin dependant. There are many modules which could call the validator, so I thought that it's not very good to declare bunch of interfaces.. The idea was to keep everything in one place..

pkg/api/api.go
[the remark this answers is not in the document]
dimkup: I think that getting panic with explicit explanation is a better user experience than unpredictable behaviour after silent truncation of input.:) I would prefer to make the check in the API.New function, set value in API struct and return it on demand, but it's another "shared state" :) Also it's not clear if serverIDFunc could return different values. (And if it cannot, why it's a function? :)

pkg/storage/limits/limits.go
insomniacslk: I am not outright against singleton factories, but I have one question about it: if the returned instance is global, we lose the ability of running tests in parallel, right? In which case this should be clearly documented at least (i.e. clarify that this is a singleton with all the disadvantages that it carries)

pkg/storage/limits/limits.go
insomniacslk: I am in favour of multiple implementations, simply because there could be multiple storage backends (where a custom plugin will help) and backend configurations (where a parametrized plugin will help)

pkg/api/api.go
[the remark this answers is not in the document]
insomniacslk: just to be clear: I am not against state (it's obviously necessary in many cases), I am wary of unnecessary global state, and this specific issue is not about global state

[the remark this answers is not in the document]
insomniacslk: `serverIDFunc` is there to allow for different implementations - I believe we use it internally to return a custom name instead of the host name (but I'm going on memory and might be wrong)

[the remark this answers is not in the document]
insomniacslk: > I think that getting panic with explicit explanation is a better user experience than unpredictable behaviour after silent truncation of input.:)

That's indeed a better user experience (and the panic is also captured and reported in the status), but what I think both @xaionaro and I are trying to say here is that we would rather have this error handled explicitly, and reported directly via error checking rather than recovering from a `panic` in the job runner

pkg/storage/limits/limits.go
dimkup: Changed to NewValidator.:) I just thought that it would be easier to switch to validation by a storage plugin in future with a single instance of validator, but let's keep it simple for now:)

pkg/storage/limits/limits.go
dimkup: Removed the interface. Probably it's better to define an interface from storage plugin to validator in future.

pkg/api/api.go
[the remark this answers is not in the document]
dimkup: Changed API.New to return error instead of panic.

dimkup:
Btw the PR does not cover TargetName and TargetID which are limited in storage as well, but I think we need a separate discussion on this topic.:)

tests/integ/jobmanager/common_go1.13.go
```diff
@@ -1,20 +0,0 @@
...
-import (
-	"errors"
-	"testing"
-
-	"github.com/stretchr/testify/require"
-)
-
-func requireErrorType(t *testing.T, err, errType error) {
-	require.True(t, errors.As(err, &errType))
-}
```
xaionaro, line 20: Why do we remove type-testing?

xaionaro (changes requested):
Requesting changes, because of downgrading the code to pre-go1.13 state in some places. I hoped we will move forward in go version (not backward). The code for go1.13 looks nice, while pre-go1.13 looks ugly, I used that approach (for go1.13) just because  I knew that some day we will declare obsolete the pre-go1.13 code. Otherwise it is a bad and ugly approach (with reflect and unit-test).

dimkup:
Sorry, removed wrong files :)

tests/integ/jobmanager/common.go
```diff
@@ -481,3 +483,7 @@ func (suite *TestJobManagerSuite) TestTestNull() {
 	_, err := suite.startJob(jobDescriptorNullTest)
 	require.Error(suite.T(), err)
 }
+
+func requireErrorType(t *testing.T, err, errType error) {
+	require.True(t, errors.As(err, &errType))
+}
```
xaionaro, line 489: This function was needed only to separate implementation for different versions of Go. Since there's no support of pre-Go1.13 anymore, this code could be used inline/inplace instead of using this interlayer.

xaionaro (approved):

pkg/storage/limits/limits_test.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/facebookincubator/contest/plugins/testfetchers/literal"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// This tests are bad, because they touche so may things which are not related to storage limitations and
+// depend on order of checks in validation code, but this is the price of having them all in one package
+
+func TestServerIDPanics(t *testing.T) {
```
xaionaro, line 33: It seems to be a wrong name of the test function :)

PR #146, opened 2020-10-05
Bugfix for amount of runs

xaionaro:
If the run start events are returns in a non-straight order then the algorithm was returning wrong amount of returns which was affecting the final reports.

### DIFF
--- a/pkg/runner/job_status.go
+++ b/pkg/runner/job_status.go
@@ -215,27 +215,30 @@ func (jr *JobRunner) BuildRunStatuses(currentJob *job.Job) ([]job.RunStatus, err
 	if err != nil {
 		return nil, fmt.Errorf("could not determine how many runs were executed: %v", err)
 	}
-	numRuns := uint(0)
 	if len(runStartEvents) == 0 {
-		return make([]job.RunStatus, 0, currentJob.Runs), nil
+		return nil, nil
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/runner/job_status.go
```diff
@@ -215,27 +215,30 @@ func (jr *JobRunner) BuildRunStatuses(currentJob *job.Job) ([]job.RunStatus, err
...
-		return make([]job.RunStatus, 0, currentJob.Runs), nil
+		return nil, nil
 	}
 
-	payload, err := runStartEvents[len(runStartEvents)-1].Payload.MarshalJSON()
-	if err != nil {
-		return nil, fmt.Errorf("could not extract JSON payload from RunStart event: %v", err)
-	}
+	numRuns := types.RunID(0)
+	for _, runStartEvent := range runStartEvents {
```
dimkup, line 223: It's probably a stupid question, but doesn't len(runStartEvents) always equal to max(RunID)? :)

xaionaro: TBH, I'm not sure (haven't checked). But my main hypothesis here is: if the job was interrupted in the middle (due to error, for example) then it could have not all runs.

But more importantly: in my understanding we just use DB as the source of trust. And we avoid multiple sources of trust to make the behavior more debuggable and predictable.

insomniacslk: I agree on the single source of truth

insomniacslk (approved):
LGTM

PR #150, opened 2020-10-12
Allow more descriptive job names with 64 chars

lestrozi:


insomniacslk:
I'd just love a comment stating where do these numbers (in the code) come from (i.e. the sql file). Other than that this looks good to me

xaionaro (approved):

insomniacslk (approved):

lestrozi:
> I'd just love a comment stating where do these numbers (in the code) come from (i.e. the sql file). Other than that this looks good to me

there's one generic comment at the top:
// Any storage is limited, so we have to be sure that data we are going to store would fit underlaying storage structures

If you want something more specific I can add it on another change if that's ok

PR #153, opened 2020-10-21
[pkg/teststeps] Avoid busy loop in ForEachTarget

insomniacslk:
I've hit a bug where cancellation leads to a busy loop until all the
plugins have returned. This happens because once the cancellation is
requested, the for-select will continuously enter the cancellation case.
With this sleep the cancellation path is not busy anymore and the CPU
doesn't suffer.
This was particuarly bad with buggy plugins that do not return quickly
enough (or at all - which is a plugin bug).

### DIFF
--- a/plugins/teststeps/teststeps.go
+++ b/plugins/teststeps/teststeps.go
@@ -7,6 +7,7 @@ package teststeps
 
 import (
 	"sync/atomic"
+	"time"
 
 	"github.com/facebookincubator/contest/pkg/cerrors"
 	"github.com/facebookincubator/contest/pkg/logging"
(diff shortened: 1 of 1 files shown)


insomniacslk:
as discussed, this is not the long-term fix but a first step. An option is to eliminate the busy loop that was introduced recently, but it's a bit more substantial as a change so I didn't take a stab at it yet

dimkup:
As a part of the review process of our previous work on ForEachTarget @xaionaro [suggested](https://github.com/facebookincubator/contest/pull/143#discussion_r486506256) another ForEachTarget implementation which were shorter and satisfying all logical requirements. (https://play.golang.org/p/Qbecxf7ThOJ)  As far as I can see, it also doesn't contain the busy loop. Could this approach be an option?

xaionaro (commented):
If this is a hotfix (not a long-term solution), then: Approve :)

xaionaro (approved):

insomniacslk:
> If this is a hotfix (not a long-term solution), then: Approve :)

I want to take a look at the other solution, it sounds a lot better than this "fix", I just have no time today unfortunately, probably tomorrow or Monday. Thank you both!

insomniacslk:
this is already addressed in another PR, closing

PR #156, opened 2020-10-27
plugins/teststeps/sshcmd: Print output to buffer and stdout

walterchris:

### DIFF
--- a/plugins/teststeps/sshcmd/sshcmd.go
+++ b/plugins/teststeps/sshcmd/sshcmd.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"os"
 	"io/ioutil"
 	"net"
 	"regexp"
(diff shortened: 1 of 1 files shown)


tfg13:
Does this mean the output gets effectively logged twice, once by piping it to stdout and then again when the command finishes and the "info stdout was: ..." logging prints it again?

walterchris:
> Does this mean the output gets effectively logged twice, once by piping it to stdout and then again when the command finishes and the "info stdout was: ..." logging prints it again?

Yeah.. I am used to gerrit nowadays - I need to clean this up. I had everything in one branch - and tried to divide it feature-wise afterwards. Looks like I did not do a good job at all.

plugins/teststeps/sshcmd/sshcmd.go
```diff
@@ -176,8 +177,10 @@ func (ts *SSHCmd) Run(cancel, pause <-chan struct{}, ch test.TestStepChannels, p
 			}
 		}()
 		// run the remote command and catch stdout/stderr
-		var stdout, stderr bytes.Buffer
-		session.Stdout, session.Stderr = &stdout, &stderr
+		var stdoutBuf, stderrBuf bytes.Buffer
+		session.Stdout = io.MultiWriter(os.Stdout, &stdoutBuf)
+		session.Stderr = io.MultiWriter(os.Stderr, &stderrBuf)
```
xaionaro, line 182: I think this is wrong to non-optionally hardcode `os.Stdout`/`os.Stderr` here. Can we add an option which enables this `os.Stdout`/`os.Stderr` injection? And add a comment why such options could be needed.

xaionaro (changes requested):

PR #157, opened 2020-10-27
plugins/teststeps/cmd: Add working dir parameter

walterchris:
Working dir parameter defines where this test steps should be executed.

### DIFF
--- a/plugins/teststeps/cmd/cmd.go
+++ b/plugins/teststeps/cmd/cmd.go
@@ -11,6 +11,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
+	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


tfg13:
Is this stacked, and github shows it in a weird way? Except for one logging line, this seems to be the same as https://github.com/facebookincubator/contest/pull/156

plugins/teststeps/cmd/cmd.go
```diff
@@ -80,14 +82,16 @@ func (ts *Cmd) Run(cancel, pause <-chan struct{}, ch test.TestStepChannels, para
 		}
 		cmd := exec.CommandContext(ctx, ts.executable, args...)
 		cmd.Dir = ts.dir
-		var stdout, stderr bytes.Buffer
-		cmd.Stdout, cmd.Stderr = &stdout, &stderr
+		var stdoutBuf, stderrBuf bytes.Buffer
+
+		cmd.Stdout = io.MultiWriter(os.Stdout, &stdoutBuf)
+		cmd.Stderr = io.MultiWriter(os.Stderr, &stderrBuf)
```
xaionaro, line 88: It does not look like "Add working dir parameter". Could you elaborate why these lines are required?

xaionaro (changes requested):

Issue #152, opened 2020-10-20
Make success optional


walterchris:
From time to time I do run commands that are allowed to fail because they do expect to fail (sometimes) - but it is nothing I can change. How do I work around these things? My current idea would be to implement an successOnFailure option which can be set within the test steps. Does that makes sense to you?

I know that I can also lower the barrier of "success rate" to work around this problem - but that's not what I want.

insomniacslk:
Can you make an example of a scenario like this?

walterchris:
In this particular case flashing via the BMC takes so long that the BMC at some point in time closes the connection down. That is meant as a failure - but that's not correct. I'd rather than hook on the serial with a timeout instead of assuming it failed.

insomniacslk:
technically you can add an `|| true` at the end of the shell command, to ignore failures, but I think this use case needs a more robust solution long term.
A possibility is that we need to send SSH `keep-alive`s to the connection, and that might be just good enough to solve your problem.

Otherwise, Tmux and nohup are good enough to start the process but not to check the status, for which something custom might help. I'm happy to brainstorm some ideas on Slack, and report back here once we have an agreement :)

PR #158, opened 2020-10-27
plugins/teststeps/sshcmd: Add Timeout

walterchris:

### DIFF
--- a/pkg/logging/logger.go
+++ b/pkg/logging/logger.go
@@ -44,8 +44,39 @@ func Disable() {
 	log.SetOutput(ioutil.Discard)
 }
 
+// Trace - Set Log Level to Trace
+func Trace() {
+	log.SetLevel(logrus.TraceLevel)
+}
+
+// Debug - Set Log Level to Debug
 func Debug() {
 	log.SetLevel(logrus.DebugLevel)
+}
+
+// Info - Set Log Level to Info
+func Info() {
+	log.SetLevel(logrus.InfoLevel)
+}
+
+// Warn - Set Log Level to Warn
+func Warn() {
+	log.SetLevel(logrus.WarnLevel)
+}
+
+// Error - Set Log Level to Error
+func Error() {
+	log.SetLevel(logrus.ErrorLevel)
+}
+
+// Fatal - Set Log Level to Fatal
+func Fatal() {
+	log.SetLevel(logrus.FatalLevel)
+}
+
+// Panic - Set Log Level to Panic
+func Panic() {
+	log.SetLevel(logrus.PanicLevel)
 }
 
 func init() {


plugins/teststeps/sshcmd/sshcmd.go
```diff
@@ -253,6 +276,9 @@ func (ts *SSHCmd) validateAndPopulate(params test.TestStepParameters) error {
 	}
 	ts.Args = params.Get("args")
 	ts.Expect = params.GetOne("expect")
+	if params.GetOne("timeout").IsEmpty() {
+		ts.Port = test.NewParam(defaultTimeoutParameter)
```
tfg13, line 280:
```suggestion
		ts.Timeout = test.NewParam(defaultTimeoutParameter)
```

plugins/teststeps/sshcmd/sshcmd.go
```diff
@@ -101,6 +104,18 @@ func (ts *SSHCmd) Run(cancel, pause <-chan struct{}, ch test.TestStepChannels, p
...
+		if err != nil {
+			return fmt.Errorf("cannot expand timeout parameter: %v", err)
+		}
+
+		timeout, err := time.ParseDuration(timeoutStr)
+		if err != nil {
+			return fmt.Errorf("cannot parse timeout paramter: %v", err)
+		}
+
+		timeTimeout := time.Now().Add(timeout)
```
tfg13, line 117: Do we really need param expansion for the timeout? If not we could do the parsing in the validate method, so we catch typos at job ingestion time.

plugins/teststeps/sshcmd/sshcmd.go
```diff
@@ -186,31 +201,39 @@ func (ts *SSHCmd) Run(cancel, pause <-chan struct{}, ch test.TestStepChannels, p
...
+			case <-cancel:
+				return session.Signal(ssh.SIGKILL)
+			case <-pause:
+				return session.Signal(ssh.SIGKILL)
+			default:
+				if time.Now().After(timeTimeout) {
+					return fmt.Errorf("timed out after %s", timeout)
+				}
+				session.Signal(ssh.Signal("CONT"))
+				time.Sleep(250 * time.Millisecond)
```
tfg13, line 235: You don't need to busy wait for this, I think the usual way to do this in go is:
```suggestion
		case <-time.After(timeTimeout):
				return fmt.Errorf("timed out after %s", timeout)
				session.Signal(ssh.Signal("SIGKILL"))
```
This should only ever be reached if the first case was not reached before.
(I think you also want to abort execution then with SIGKILL)

plugins/teststeps/sshcmd/sshcmd.go
```diff
@@ -186,31 +201,39 @@ func (ts *SSHCmd) Run(cancel, pause <-chan struct{}, ch test.TestStepChannels, p
...
+				if err == nil {
+					// Execute expectations
+					expect := ts.Expect.String()
+					if expect == "" {
+						log.Warningf("no expectations specified")
 					} else {
-						return fmt.Errorf("match for %s not found for target %v", expect, target)
+						re := regexp.MustCompile(expect)
+						matches := re.FindAll(stdoutBuf.Bytes(), -1)
+						if len(matches) > 0 {
```
tfg13, line 216: what is this `stdoutBuf` stuff doing here, I assume it leaked from one of the other PRs?

tfg13 (changes requested):

plugins/teststeps/sshcmd/sshcmd.go
walterchris: I know that we do not need busy waiting for this. This actually comes from https://github.com/facebookincubator/contest/issues/152 - I had the problem that the other connection closed the connection. Turned out that I constantly need to fire Signals in order to keep the connection alive. I tried implementing keep alive which did not help - so I just this.

plugins/teststeps/sshcmd/sshcmd.go
walterchris: Yes that should not be here. Let me fix this.

plugins/teststeps/sshcmd/sshcmd.go
tfg13: Aaaah, that is what the CONT is for. Please drop a comment in then, so people know :)

tfg13: Wait, but why does this help against the connection being closed explicitly? Sry, not sure I follow :/
And 4 signals per second seems like a lot :D

PR #159, opened 2020-10-28
feature(api): Add option to set event timeout

xaionaro:
# Description

~~Currently we return an error if an API request wasn't processed in 3 seconds. Which appears quite often (by undiagnosed reason), so we temporary increase the timeout to give us time to find a proper solution. In logs I see that it took 4 seconds to process the request. So I'm raising the value to 10 seconds.~~

Before this fix processing time in API was hardcoded to 3 seconds, which is not always a good setting. To make this setting changable, we introduce here a special option.
    
Also since we now have options for API, we moved `serverIDFunc` to it.

---

# Test Plan
```
xaionaro@void:~/go/src/github.com/facebookincubator/contest$ go test ./...
?       github.com/facebookincubator/contest/cmds/clients/contestcli-http       [no test files]
?       github.com/facebookincubator/contest/cmds/contest       [no test files]
ok      github.com/facebookincubator/contest/pkg/api    (cached)
?       github.com/facebookincubator/contest/pkg/cerrors        [no test files]
?       github.com/facebookincubator/contest/pkg/config [no test files]
?       github.com/facebookincubator/contest/pkg/event  [no test files]
?       github.com/facebookincubator/contest/pkg/event/frameworkevent   [no test files]
?       github.com/facebookincubator/contest/pkg/event/internal/querytools      [no test files]
?       github.com/facebookincubator/contest/pkg/event/internal/reflecttools    [no test files]
ok      github.com/facebookincubator/contest/pkg/event/testevent        (cached)
?       github.com/facebookincubator/contest/pkg/job    [no test files]
?       github.com/facebookincubator/contest/pkg/jobmanager     [no test files]
ok      github.com/facebookincubator/contest/pkg/lib/comparison (cached)
?       github.com/facebookincubator/contest/pkg/logging        [no test files]
ok      github.com/facebookincubator/contest/pkg/pluginregistry (cached)
ok      github.com/facebookincubator/contest/pkg/runner (cached)
?       github.com/facebookincubator/contest/pkg/storage        [no test files]
ok      github.com/facebookincubator/contest/pkg/storage/limits 0.003s
ok      github.com/facebookincubator/contest/pkg/target (cached)
ok      github.com/facebookincubator/contest/pkg/test   (cached)
?       github.com/facebookincubator/contest/pkg/types  [no test files]
?       github.com/facebookincubator/contest/plugins/listeners/httplistener     [no test files]
?       github.com/facebookincubator/contest/plugins/reporters/noop     [no test files]
?       github.com/facebookincubator/contest/plugins/reporters/targetsuccess    [no test files]
ok      github.com/facebookincubator/contest/plugins/storage/memory     (cached)
?       github.com/facebookincubator/contest/plugins/storage/rdbms      [no test files]
?       github.com/facebookincubator/contest/plugins/targetlocker/dblocker      [no test files]
ok      github.com/facebookincubator/contest/plugins/targetlocker/inmemory      (cached)
ok      github.com/facebookincubator/contest/plugins/targetlocker/noop  (cached)
?       github.com/facebookincubator/contest/plugins/targetmanagers/csvtargetmanager    [no test files]
?       github.com/facebookincubator/contest/plugins/targetmanagers/targetlist  [no test files]
?       github.com/facebookincubator/contest/plugins/testfetchers/literal       [no test files]
?       github.com/facebookincubator/contest/plugins/testfetchers/uri   [no test files]
ok      github.com/facebookincubator/contest/plugins/teststeps  (cached)
?       github.com/facebookincubator/contest/plugins/teststeps/cmd      [no test files]
?       github.com/facebookincubator/contest/plugins/teststeps/echo     [no test files]
?       github.com/facebookincubator/contest/plugins/teststeps/example  [no test files]
?       github.com/facebookincubator/contest/plugins/teststeps/randecho [no test files]
?       github.com/facebookincubator/contest/plugins/teststeps/slowecho [no test files]
?       github.com/facebookincubator/contest/plugins/teststeps/sshcmd   [no test files]
?       github.com/facebookincubator/contest/plugins/teststeps/terminalexpect   [no test files]
?       github.com/facebookincubator/contest/tests/plugins/teststeps/channels   [no test files]
?       github.com/facebookincubator/contest/tests/plugins/teststeps/crash      [no test files]
?       github.com/facebookincubator/contest/tests/plugins/teststeps/fail       [no test files]
?       github.com/facebookincubator/contest/tests/plugins/teststeps/hanging    [no test files]
?       github.com/facebookincubator/contest/tests/plugins/teststeps/noop       [no test files]
?       github.com/facebookincubator/contest/tests/plugins/teststeps/noreturn   [no test files]
?       github.com/facebookincubator/contest/tests/plugins/teststeps/panicstep  [no test files]
```


insomniacslk (commented):
I'd suggest to change `api.New` to accept a timeout parameter rather than changing a default value (given that the intention is to temporarily bump this value on our side, rather than using a more reasonable default).

The `api.API` object is created in `JobManager.Start`, so we should expose the API timeout value there too. I am wondering if we shouldn't just change `JobManager.Start` to get an `*api.API` object, and create it only if it's not nil

xaionaro:
> The `api.API` object is created in `JobManager.Start`, so we should expose the API timeout value there too. I am wondering if we shouldn't just change `JobManager.Start` to get an `*api.API` object, and create it only if it's not nil

I like this. An I think I would go even further: we can create `apiListener` outside of `JobManager` as well. This way `JobManager` will also require only `chan *Event` which reduces boundness/dependencyness. But meanwhile I'll propose a less radical solution in this PR: I'll just add `Option`-s to `Start` :)

xaionaro:
@insomniacslk : I've just updated the approach.

pkg/api/options.go
```diff
@@ -0,0 +1,54 @@
+package api
+
+import (
+	"time"
+)
+
+// Option is an additional argument to method New to change the behavior
+// of API processing.
+type Option interface {
```
insomniacslk, line 9: Nit: this could probably be simply a function, e.g. `type Option func(*Config)` (but this is fine too, especially if you foresee to add more methods)

pkg/jobmanager/options.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+// OptionEventTimeout defines time duration for API request to be processed.
+// If a request is not processed in time, then an error is returned to the
+// client.
+type OptionAPI struct {
+	api.Option
+}
+
+// Apply implements Option.
```
insomniacslk, line 24: Function name and comment are not matching

pkg/jobmanager/options.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+type config struct {
+	apiOptions []api.Option
+}
+
+// OptionEventTimeout defines time duration for API request to be processed.
+// If a request is not processed in time, then an error is returned to the
+// client.
+type OptionAPI struct {
```
insomniacslk, line 20: Structure name and comment are not matching

insomniacslk (approved):

pkg/api/options.go
xaionaro: I do not like dynamically created functions passed through packages. For example, I cannot compare functions, serialize them or at least to print its value. I understand all these is not used. I don't mind to change it to function-based approach if you still want to. But until then I would like to preserve current approach :)

pkg/jobmanager/options.go
xaionaro: Oh, I pushed a corrupted code. It should've been `apply` (not `a`). Fixed. Thank you!

xaionaro:
Rebased to master and fixed to pass CI-tests (license headers and a fix in integration tests)

PR #128, opened 2020-08-05
TargetLocker: remove CheckLocks, enforce locking during target acquisition

insomniacslk:
CheckLocks is no longer needed, and can be misleading. I have removed
this method entirely, and enforced that all targets are locked during
the target acquisition phase.

Signed-off-by: Andrea Barberio <insomniac@slackware.it>

PR #161, opened 2020-10-30
Inmemorycleanup

tfg13:
This will make the upcoming PRs for the db locking API shorter and nicer.

The `checkLocks` call was removed in #128, but missed here I think :)

Test Plan:
```
contest % go test  ./plugins/targetlocker/inmemory
ok  	github.com/facebookincubator/contest/plugins/targetlocker/inmemory	0.357s
```


tfg13:
One sec, diff stacking is wrong

tfg13:
All good now

insomniacslk (approved):

PR #162, opened 2020-10-30
targetlocker: Add TryLock API call, implement in plugins

tfg13:
This adds `TryLock`, which can be used to batch locking calls properly. It does not error when some locks can't be acquired, and returns a list of target IDs that worked.

Test Plan:
Each of the implementations in noop, inmemory, dblocker is separate and covered by their own (new) unit tests.
Also running the integration suite. The new API called is not used yet of course, future commits will do that.

### DIFF
--- a/pkg/target/locker.go
+++ b/pkg/target/locker.go
@@ -34,6 +34,14 @@ type Locker interface {
 	// Locks are reentrant, locking existing locks (with the same owner)
 	// extends the deadline.
 	Lock(types.JobID, []*Target) error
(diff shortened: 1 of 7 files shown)


tfg13:
This contains #160, will rebase when that one is merged to make this more readable

tfg13:
This is rebased and ready for review now

xaionaro (approved):
LGTM. Haven't check unit-tests, but I do not think there could be any important issue.

PR #163, opened 2020-11-02
eventemitter: Enforce TestStep can only emit registered events

tfg13:
This was always intended, but actually never enforce when emitting events.

It is enforced, however, when loading old test steps/events from the database, which leads to the situation where emitting unregistered events works, but then running `status` against such a job does not.

### DIFF
--- a/pkg/storage/events.go
+++ b/pkg/storage/events.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/facebookincubator/contest/pkg/event"
 	"github.com/facebookincubator/contest/pkg/event/frameworkevent"
 	"github.com/facebookincubator/contest/pkg/event/testevent"
 )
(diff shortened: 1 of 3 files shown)


pkg/storage/events.go
```diff
@@ -30,6 +33,11 @@ type TestEventEmitterFetcher struct {
 
 // Emit emits an event using the selected storage layer
 func (e TestEventEmitter) Emit(data testevent.Data) error {
+	if e.allowedEvents != nil {
+		if _, ok := (*e.allowedEvents)[data.EventName]; !ok {
```
insomniacslk, line 37: does this mean that if no event is specified in the plugin, any event emission will be permitted?

tfg13: yes, this is for contest itself, it emits things like `EventTestError`, which are not specifically registered

insomniacslk: should we register them instead (or track this in an issue)? This feels dangerous for plugins that don't declare events but emit them

insomniacslk (approved):
Approving, this is already an improvement over what we have now

pkg/storage/events.go
tfg13: Why? The plugin has no choice, it gets an event emitter from the pipeline that has this set and it can't change that, no way to get around. And the pipeline has its own emitter without restrictions

tfg13: (at least that is what I intended)

PR #173, opened 2020-11-16
plugins/teststeps/sshcmd: Add Timeout and busy waiting

walterchris:
1. Add Timeout to the command
2. Add busy waiting. Busy waiting is needed to catch the expect
   parameter. Otherwise, if we have a continuous command e.g. reading
   from the command line it will never complete successfully, thus it
   will also not be parsed for the expect parameter - and thus never
   success.

I added a "dummy" CONT to keep the ssh connection busy - such that it
does not close down because of a time out.

### DIFF
--- a/plugins/teststeps/sshcmd/sshcmd.go
+++ b/plugins/teststeps/sshcmd/sshcmd.go
@@ -25,6 +25,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/facebookincubator/contest/pkg/cerrors"
 	"github.com/facebookincubator/contest/pkg/event"
@@ -48,6 +49,7 @@ var log = logging.GetLogger("teststeps/" + strings.ToLower(Name))
 var Events = []event.Name{}
 
 const defaultSSHPort = 22
+const defaultTimeoutParameter = "10m"
 
 // SSHCmd is used to run arbitrary commands as test steps.
 type SSHCmd struct {
@@ -59,6 +61,7 @@ type SSHCmd struct {
 	Executable     *test.Param
 	Args           []test.Param
 	Expect         *test.Param
+	Timeout        *test.Param
 }
 
 // Name returns the plugin name.
@@ -100,6 +103,18 @@ func (ts *SSHCmd) Run(cancel, pause <-chan struct{}, ch test.TestStepChannels, p
 		if err != nil {
 			return fmt.Errorf("failed to convert port parameter to integer: %v", err)
 		}
+
+		timeoutStr, err := ts.Timeout.Expand(target)
+		if err != nil {
+			return fmt.Errorf("cannot expand timeout parameter %s: %v", timeoutStr, err)
+		}
+
+		timeout, err := time.ParseDuration(timeoutStr)
+		if err != nil {
+			return fmt.Errorf("cannot parse timeout paramter: %v", err)
+		}
+
+		timeTimeout := time.Now().Add(timeout)
 
 		// apply functions to the private key, if any
 		var signer ssh.Signer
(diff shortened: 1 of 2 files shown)


plugins/teststeps/sshcmd/sshcmd.go
```diff
@@ -186,31 +201,49 @@ func (ts *SSHCmd) Run(cancel, pause <-chan struct{}, ch test.TestStepChannels, p
...
+					return fmt.Errorf("timed out after %s", timeout)
+				}
+				err = session.Signal(ssh.Signal("CONT"))
+				if err != nil {
+					log.Warnf("Unable to send CONT to ssh server: %v", err)
 				}
-			} else {
-				log.Warningf("Stderr of command '%s' is '%s'", cmd, stderr.Bytes())
+				// Sanity Break to not spam the Server
+				time.Sleep(250 * time.Millisecond)
```
rojer, line 245: you can select on `<- time.After(250 * time.Millisecond)` instead

plugins/teststeps/sshcmd/sshcmd.go
```diff
@@ -186,31 +201,49 @@ func (ts *SSHCmd) Run(cancel, pause <-chan struct{}, ch test.TestStepChannels, p
...
+				re := regexp.MustCompile(expect)
+				matches := re.FindAll(stdout.Bytes(), -1)
+				if len(matches) > 0 {
+					log.Infof("match for regex \"%s\" found", expect)
+					return nil
+				}
+				if time.Now().After(timeTimeout) {
+					return fmt.Errorf("timed out after %s", timeout)
+				}
+				err = session.Signal(ssh.Signal("CONT"))
```
rojer, line 240: can you add a comment on why this is necessary?

plugins/teststeps/sshcmd/sshcmd.go
```diff
@@ -186,31 +201,49 @@ func (ts *SSHCmd) Run(cancel, pause <-chan struct{}, ch test.TestStepChannels, p
...
+				} else {
+					log.Warningf("Stderr of command '%s' is '%s'", cmd, stderr.Bytes())
+				}
+				return err
+			case <-cancel:
+				return session.Signal(ssh.SIGKILL)
+			case <-pause:
+				return session.Signal(ssh.SIGKILL)
+			default:
+				re := regexp.MustCompile(expect)
```
rojer, line 231: compile this once, after retrieving.
also, assuming this is supplied by the user, this will crash in case of a bad regex, which is bad.
MustCompile should only be used for static regexes.

plugins/teststeps/sshcmd/sshcmd.go
```diff
@@ -186,31 +201,49 @@ func (ts *SSHCmd) Run(cancel, pause <-chan struct{}, ch test.TestStepChannels, p
...
+				matches := re.FindAll(stdout.Bytes(), -1)
+				if len(matches) > 0 {
+					log.Infof("match for regex \"%s\" found", expect)
```
rojer, line 234: avoid escaping by using ticks. or change to single quotes (as is done in other places).

rojer (changes requested):

plugins/teststeps/sshcmd/sshcmd.go
walterchris: Ack

plugins/teststeps/sshcmd/sshcmd.go
walterchris: Ack

plugins/teststeps/sshcmd/sshcmd.go
walterchris: Done.

plugins/teststeps/sshcmd/sshcmd.go
walterchris: Done.

plugins/teststeps/sshcmd/sshcmd.go
rojer: this will generate quite a bit of activity on the connection, we'll be sending and receiving 4 packets every second.
do you know why the connection gets dropped? is it inactivity or what?

walterchris: Not sure.

I tried working around this with ServerAliveInterval - The BMC (OpenBMC) but actually requires "keystrokes" to not shut down the connection to the client. Thus I'm sending CONT to keep the connection. IMHO 4 CONT's a second isn't too much noise - but I do see the point.

We can work around this by skipping this - or only sending CONT every x times. Otherwise I can also leave it out and have it sitting in my own fork until we worked around this.

walterchris: I can also guard this with a flag - if that's better.

rojer: let's do every 20th time, this way we only send every 5 seconds

walterchris: Done.

rojer: you forgot to ++ the counter

walterchris: :D - it's 01:49am..

walterchris: Done. Forced-Push - never happend.

rojer: :)

PR #174, opened 2020-11-16
[rdbms][migration] Make extended_descriptor migration idempotent

marcoguerri:
Adding a column in an idempotent way and not making assumption
on the backend and the SQL dialect is not easy. So, move the
extended_descriptor migration to pure Go, which makes it easier
to implement the addition in an idempotent way.

### DIFF
--- a/db/rdbms/migration/0001_add_extended_descriptor.go
+++ b/db/rdbms/migration/0001_add_extended_descriptor.go
@@ -1,0 +1,48 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+//
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+package migration
+
+import (
+    "database/sql"
+
+    "github.com/facebookincubator/contest/tools/migration/rdbms/migrate"
+    "github.com/go-sql-driver/mysql"
+    "github.com/sirupsen/logrus"
+)
+
+type ExtendedDescriptorMigration struct {
+    log *logrus.Entry
+}
+
+// Up implements the forward migration
+func (m *ExtendedDescriptorMigration) Up(tx *sql.Tx) error {
+    _, err := tx.Query("alter table jobs add column extended_descriptor text")
+    if err != nil {
+        if mysqlErr, ok := err.(*mysql.MySQLError); ok {
+            if mysqlErr.Number == 1060 {
+                m.log.Warningf("extended_descriptor column already exists, migration will have no effect")
+                return nil
+            }
+        }
+        return err
+    }
+    return nil
+}
+
+// Down implements the backward migration
+func (m *ExtendedDescriptorMigration) Down(tx *sql.Tx) error {
+    return nil
+}
+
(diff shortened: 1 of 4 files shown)


db/rdbms/migration/0001_add_extended_descriptor.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// Down implements the backward migration
+func (m *ExtendedDescriptorMigration) Down(tx *sql.Tx) error {
+    return nil
+}
+
+// NewExtendedDescriptorMigration is the factory for ExtendedDescriptorMigration
+func NewExtendedDescriptorMigration(log *logrus.Entry) migrate.Migrate {
+    return &ExtendedDescriptorMigration{log: log}
```
xaionaro, line 42: **nitpicking:**

It would be nice to set `log` to something dummy-like if a `nil` was passed, IMHO.

xaionaro (approved):
I do not understand why is that required, since we control schema revision versions (I hope we do so, at least), anyway. So we should not do the same migration twice, ever. If it happened, then somethings is wrong. So there should be no problem to report a problem if we try to roll-up a migration, which is already rolled-up.

db/rdbms/migration/0001_add_extended_descriptor.go
marcoguerri: > I do not understand why is that required, since we control schema revision versions (I hope we do so, at least), anyway. So we should not do the same migration twice, ever.

The schema version is tracked, the same migration will not run twice. But internally schema migrations are not going to be controlled by this tool (we have a different infra for schema migrations). Other migrations, e.g. ([0002_migrate_descriptor_to_extended_descriptor.go](https://github.com/facebookincubator/contest/blob/master/db/rdbms/migration/0002_migrate_descriptor_to_extended_descriptor.go)) will be under the control of this tool. 

Therefore, I still need for them to be idempotent, the first time the migration runs and catches up with the schema version. Otherwise internally I will need to separate `sql` migrations from `go` migrations. I would instead like to keep them in a single please and always refer to the upstream set of migration as source of truth.

xaionaro: Sorry, I do not follow:

> Other migrations, e.g. (0002_migrate_descriptor_to_extended_descriptor.go) will instead be under the control of this tool.

"this tool" is the upstream one or the internal one?

> Therefore, I still need for them to be idempotent, the first time the migration runs and catches up with the schema version.

Why not to fix the number in the internal DB, manually?

> Otherwise internally I will need to separate `sql` migrations from `go` migrations.

Why? :)

---

What we can do is separate schema migrations from data migrations. Which is a quite normal practice, IMO. But this requires more time investment... Anyway, do not waste time on the answers: I do not see negative impact from this PR, so let's just merge it :)

marcoguerri: I will separate `sql` migrations from `go` migrations internally. I will resume this later if the manual operations turn out to be burden.

PR #178, opened 2020-11-20
targetlocker: Add limit parameter to TryLock

tfg13:
`TryLock` currently attempts to lock as many targets as possible. This works, but in many cases targetmanagers want to hit a specific number of targets in the end, and since they can't predict which targets will fail locking, they provide more candidates to `TryLock`.

This can overshoot the target, locking more than the user really needs. Targetmanagers can work around this by calling UnLock at the cost of another roundtrip to the lock service/database.

This PR adds a `limit` parameter to `TryLock` and implements it in all three lockers (noop, inmemory, dblocker). The lockers will then internally try to acquire locks until they reach the limit and stop.
Note this is strictly just an upper limit, it is not an error to lock less targets than the limit says (or even zero targets).

Test Plan:
This should be fully covered by new and old unit tests.

### DIFF
--- a/pkg/target/locker.go
+++ b/pkg/target/locker.go
@@ -34,14 +34,14 @@ type Locker interface {
 	// Locks are reentrant, locking existing locks (with the same owner)
 	// extends the deadline.
 	Lock(types.JobID, []*Target) error
-	// TryLock attempts to lock all given targets.
+	// TryLock attempts to lock up to limit of the given targets.
 	// The job ID is the owner of the lock.
-	// This function attempts to lock as many of the given targets as possible,
+	// This function attempts to lock up to limit of the given targets,
 	// and returns a list of target IDs that it was able to lock.
 	// This function does not return an error if it was not able to lock any targets.
 	// Locks are reentrant, locking existing locks (with the same owner)
 	// extends the deadline.
-	TryLock(types.JobID, []*Target) ([]string, error)
+	TryLock(jobID types.JobID, targets []*Target, limit uint) ([]string, error)
 	// Unlock unlocks the specificied targets if they are held by the given owner.
 	// Unlock silently skips expired locks and targets that are not locked at all.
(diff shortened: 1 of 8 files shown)


pkg/target/locker.go
```diff
@@ -34,14 +34,14 @@ type Locker interface {
...
+	// TryLock attempts to lock up to limit of the given targets.
 	// The job ID is the owner of the lock.
-	// This function attempts to lock as many of the given targets as possible,
+	// This function attempts to lock up to limit of the given targets,
 	// and returns a list of target IDs that it was able to lock.
 	// This function does not return an error if it was not able to lock any targets.
 	// Locks are reentrant, locking existing locks (with the same owner)
 	// extends the deadline.
-	TryLock(types.JobID, []*Target) ([]string, error)
+	TryLock(jobID types.JobID, targets []*Target, limit int) ([]string, error)
```
insomniacslk, line 44: could this be `uint`? In which case you can also get rid of the `<= 0` check below that every plugin will have to implement (the case when it's 0 is still up though)

insomniacslk (approved):

pkg/target/locker.go
tfg13: Switched to `uint` because people seem to like it better.

I still disagree :)
The API does indeed look nicer and it makes the purpose of the parameter more clear, but now it is more cumbersome for _both_ the user and the implementer as one needs to cast to `uint` all the time, since builtins like `len()` return `int`.

I don't think any type-safety argument can be made here, as it will happily cast negative values to huge integers without notice at runtime, my view is one might as well accept `int` and implement some sane behaviour (or clear exception) for negative values.

Note this change to `uint` creates a subtile change in behavior. Before, plugins would not lock any targets on negative values. Now an accidentally casted negative value is a huge integer, so the plugin will lock many/all targets.

PR #182, opened 2020-11-24
pkg/target: Rework target to have unique ID, remove name

tfg13:
### tldr
This changes the fields in the core `Target` struct from `Name, ID, FQDN` to `ID, FQDN, IPv4, IPv6`, with only `ID` required.
As this is a **major, breaking change**, feedback is appreciated.

### Rationale
This aims to solve the following problems with the way targets work at the moment:
- `Target` currently has 3 fields (`Name, ID, FQDN`) without clear rules what should be set and what is optional. This poses a problem for storage/database plugins as there is no obvious primary key. This is addressed by explicitly marking `ID` as the primary target identifier, which is now mandatory and must be unique.
- No well-defined way to store IP addresses. We anticipate most targets to be somehow addressed/controlled/tested via the network. At the moment, the IP might be in either `Name or ID`. This is addressed by explicitly adding `PrimaryIPv6` and `PrimaryIPv4`. We expect this field to be filled most of the time, but it is optional to support environments where DNS lookups are used or targets are not contacted via the IP network.
- Plugins (not targetmanagers) sometimes have to (re)construct `Target` objects for return values and similar. If they know one identifier (like the `ID`), it is unclear if they need to resolve and fill the other values before returning `Target`s. This is partially addressed by explicitly only making `ID`s mandatory. This is not a complete fix on its own, but it allows for one. The new `dblocker` (plugins/targetlocker) for example accepts `Target`s as input, but returns strictly only (`string`) `ID`s
- Comparing targets (and indexing maps with targets) is problematic. Targets might come from different sources (target managers, storages) with no strict guarantees if all the optional fields are filled, leading to surprising equality check results. This is addressed indirectly by adding `net.IP` to `Target`. This type is not comparable, making `Target` comparisons (and indexing) a compile-time error. In most cases this can be trivially replaced by comparing and indexing `Target.ID`

### Scenarios
Here are some environments we try to cover with this rework:
- Large places with inventory management systems. These systems often assign some unique ID to devices. Users can set `ID` to their internal inventory IDs if this is a useful way for lookups, and still have `FQDN` and the `IP` fields free for addressing.
- Hostnames as unique IDs. Either short names or fully qualified. Users should put their hostnames in both the `ID` and `FQDN` fields, and fill the `IP` fields if possible/useful.
- IP addresses as primary IDs. Users should put a normalised(!) textual representation into the `ID` field and also fill the matching `IP` field. `FQDN` can be filled as required.
- Not IP-addressable test devices. We expect this to be rare, but it is supported. If whatever serves as address is unique, it can be put it in `ID`. Since many built-in plugins like `SSH` can't be used anyway, the `FQDN` field can be used as a secondary string if absolutely necessary.

**If you have a usecase that is not covered, please reach out!**

### Not addressed here
- There was, and still is, ambiguity on how to contact a device on the network. Plugins must choose between `FQDN`, `PrimaryIPv6` and `PrimaryIPv4`. Note this problem existed before as well. Ideally, generic tests steps like SSH support all of them, while custom plugins are free to only support one/some, as they are only used in specific environments.

### Known breakage caused by this change
- `CSVFileTargetManager`: Format changed from `Name, ID` to `ID, FQDN, IPv6, IPv4`
- `TargetList`: This unmarshals `Target` structs from acquire parameters, they need to be updated (`Name, ID, FQDN` to `ID, Name, PrimaryIPv6, PrimaryIPv4`)
- Some custom `TestStep` implementations that use `Name`
- `Target` can no longer be compared, might break status struct comparisons with embedded `Targets` in some clients (TODO investigate)

### Open questions (*cough* feedback *cough*)
- `test_events` in the main storage (rdbms) plugin currently stores `target_id` and `target_name`. Should this be normalised to only store `target_id`. Does this need a new `targets` table for full event reconstruction for job resumption? (leaning towards only store id in this PR, introduce the targets table with another one related to resumption)
- `Target.ID` is of type `string`. Is this ok or should we declare a custom type?
- `Target.FQDN` is of type `*string`. The idea was to have it point to `ID` for environments where FQDNs are used as primary IDs. However, this is the only pointer in `Target`, assigning it inline can be a bit cumbersome. Should it just be `string` as well, as the memory impact will be minimal? (Leaning towards yes, make it `string`)
- (Semi-related, only mentioned because we are already refactoring this entire thing) Storage plugins will often enforce some length limit on the `ID`, `FQDN` fields. There is currently no way to express/validate targets before going to the storage layer.

### DIFF
--- a/pkg/test/parameter_test.go
+++ b/pkg/test/parameter_test.go
@@ -16,37 +16,37 @@ import (
 
 func TestParameterExpand(t *testing.T) {
 	validExprs := [][4]string{
-		// expression, target name, target ID, expected result
-		[4]string{"{{ ToLower .Name }}", "www.slackware.IT", "1234", "www.slackware.it"},
-		[4]string{"{{ .Name }}", "www.slackware.IT", "2345", "www.slackware.IT"},
-		[4]string{"{{ .Name }}", "www.slackware.it", "3456", "www.slackware.it"},
-		[4]string{"name={{ .Name }}, id={{ .ID }}", "blah", "12345", "name=blah, id=12345"},
+		// expression, target FQDN, target ID, expected result
+		[4]string{"{{ ToLower .FQDN }}", "www.slackware.IT", "1234", "www.slackware.it"},
+		[4]string{"{{ .FQDN }}", "www.slackware.IT", "2345", "www.slackware.IT"},
+		[4]string{"{{ .FQDN }}", "www.slackware.it", "3456", "www.slackware.it"},
+		[4]string{"fqdn={{ .FQDN }}, id={{ .ID }}", "blah", "12345", "fqdn=blah, id=12345"},
 	}
 	for _, x := range validExprs {
 		p := NewParam(x[0])
-		res, err := p.Expand(&target.Target{Name: x[1], ID: x[2]})
+		res, err := p.Expand(&target.Target{FQDN: x[1], ID: x[2]})
 		require.NoError(t, err, x[0])
 		require.Equal(t, x[3], res, x[0])
 	}
 }
 
 func TestParameterExpandUserFunctions(t *testing.T) {
-	title := func(a ...string) (string, error) {
+	customFunc := func(a ...string) (string, error) {
 		if len(a) == 0 {
 			return "", errors.New("no params")
 		}
 		return strings.Title(a[0]), nil
 	}
-	err := RegisterFunction("title", title)
+	err := RegisterFunction("CustomFunc", customFunc)
 	require.NoError(t, err)
 	validExprs := [][4]string{
-		// expression, target name, target ID, expected result
-		[4]string{"{{ Title .Name }}", "slackware.it", "1234", "Slackware.It"},
-		[4]string{"{{ Title .ID }}", "slackware.it", "a1234a", "A1234a"},
+		// expression, target FQDN, target ID, expected result
+		[4]string{"{{ CustomFunc .FQDN }}", "slackware.it", "1234", "Slackware.It"},
+		[4]string{"{{ CustomFunc .ID }}", "slackware.it", "a1234a", "A1234a"},
 	}
 	for _, x := range validExprs {
 		p := NewParam(x[0])
-		res, err := p.Expand(&target.Target{Name: x[1], ID: x[2]})
+		res, err := p.Expand(&target.Target{FQDN: x[1], ID: x[2]})
 		require.NoError(t, err, x[0])
 		require.Equal(t, x[3], res, x[0])
 	}
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ moment).
 
 ## Quick Start
 
-We offer a few Docker files to help setup ConTest. The fastest way to 
+We offer a few Docker files to help setup ConTest. The fastest way to
 have a ConTest instance up and running is to bring up the server and MySQL
 containers via the docker-compose configuration: just run
 `docker-compose up --build` from the root of the source tree.
@@ -348,7 +348,7 @@ commented for clarity:
             // on the actual plugin.
             "TargetManagerAcquireParameters": {
                 // the URI for the CSV file that contains a list of targets in
-                the format "ip,hostname". This is intentionally very simple.
+                the format "id,fqdn,ipv4,ipv6". This is intentionally very simple.
                 "FileURI": "hosts02.csv",
                 // The minimum number of targets needed for the test. If we
                 // don't get at least this number of devices, the job will fail.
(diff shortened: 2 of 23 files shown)


xaionaro:
First comments:

1. I do not understand why do we need both `IPv4` and `IPv6`. Why not to store one IP address (basically just `net.IP`, which could be as v4, as v6)?
2. In ideal world I would prefer to do not store any IP address here. Instead I would prefer to make pluggable transport dialers. Like [http.Transport.Dial](https://golang.org/pkg/net/http/#Transport).

xaionaro:
Addressing a couple of "Open questions":

> * `Target.ID` is of type `string`. Is this ok or should we declare a custom type?

I think we should declare a type. It could be even just `type ID string`. This way the code will guarantee more safety through type-checking, since target.ID is a crucial part of a lot of logic.

> * `Target.FQDN` is of type `*string`. The idea was to have it point to `ID` for environments where FQDNs are used as primary IDs. However, this is the only pointer in `Target`, assigning it inline can be a bit cumbersome. Should it just be `string` as well, as the memory impact will be minimal? (Leaning towards yes, make it `string`)

What is the reason of the idea to point to `ID`? The memory impact is about zero, since strings are internally read-only variables are assigned by pointer (not by contents), anyway.

xaionaro (approved):
In total, IMO, it became better than it was: a lot of vagueness/implicitness is removed (for example now it is clear how to compare targets).

tfg13:
Change:
- Made FQDN `string`, not `*string` as it is nicer to work with
- Introduced `types.TargetID`. This is nice type-safety wise, but stumbled upon an unfortunate consequence that might break it a little, see the comment on `parameter_test.go`

pkg/test/parameter_test.go
```diff
@@ -7,46 +7,54 @@ package test
...
-	err := RegisterFunction("title", title)
+	err := RegisterFunction("CustomFunc", customFunc)
 	require.NoError(t, err)
 	validExprs := [][4]string{
-		// expression, target name, target ID, expected result
-		[4]string{"{{ Title .Name }}", "slackware.it", "1234", "Slackware.It"},
-		[4]string{"{{ Title .ID }}", "slackware.it", "a1234a", "A1234a"},
+		// expression, target FQDN, target ID, expected result
+		[4]string{"{{ CustomFunc .FQDN }}", "slackware.it", "1234", "Slackware.It"},
+		[4]string{"{{ CustomFunc .ID }}", "slackware.it", "a1234a", "A1234a"},
```
tfg13, line 53: We can't use a `func(a ...string) (string, error)` anymore if we want to template `func .ID`. The template parsing/execution doesn't understand and breaks:
```
--- FAIL: TestParameterExpandUserFunctions (0.00s)
    parameter_test.go:58:
        	Error Trace:	parameter_test.go:53
        	Error:      	Received unexpected error:
        	            	template: :1:9: executing "" at <.ID>: wrong type for value; expected string; got types.TargetID
        	Test:       	TestParameterExpandUserFunctions
        	Messages:   	{{ Title .ID }}
FAIL
```
The workaround is accepting `interface` and doing a type switch, but this is ugly and means we can no longer use simple things like `strings.ToUpper`

insomniacslk:
Catching up now as requested - I see that some questions have been answered already. On making `FQDN` a `string` instead of `*string` I agree. Memory-wise it is fine as Dmitrii suggested, and the only reason to use a pointer would be to signal its absence, but we can unambiguously use the empty string for that purpose. 

On comparing targets - they can still be compared by ID, but if we want to make this explicit, we can add a `Target.Equals` method.

On how to store targets in test events - one of the reasons behind this change was to allow serialization of targets for that purpose. I would like to keep that in, and have a `String` or similar method (and maybe a `Parse` counterpart to reconstruct it? Or maybe not) that allows to serialize and store targets without information loss.

On the storage limits on the `ID` string, I guess we can live with what we have currently, and tackle it in a separate PR.

pkg/test/parameter_test.go
insomniacslk: I am for reverting from `types.TargetID` to `string`, it gives us more trouble than benefits

tfg13:
I reverted the `types.TargetID` change, it makes templating very ugly

tfg13:
Running some more tests locally, but this should be ready for review

tfg13:
I ran a job with the old version, then applied these changes, then queried its status. It still works because json unmarshalling is pretty forgiving, but the contents of the `Name` field are lost. I hope this is acceptable.

tfg13:
Updated all the examples I could fine. Comment: They are a bit of a mess.

PR #184, opened 2020-11-27
[pkg/teststeps] Implement ForEachNonResumingTarget helper function and use it in cmd plugin

rihter007:
Introduce ForEachNonResumingTarget helper function that provides easier functionality for plugins that do not support pause/restore.

It is also useful if plugin internally uses a blocking operation that takes context.Context as its input
Rewrite cmd plugin using the new ForEachNonResumingTarget

### DIFF
--- a/plugins/teststeps/teststeps.go
+++ b/plugins/teststeps/teststeps.go
@@ -6,6 +6,7 @@
 package teststeps
 
 import (
+	"context"
 	"sync/atomic"
 
 	"github.com/facebookincubator/contest/pkg/cerrors"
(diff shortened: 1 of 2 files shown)


plugins/teststeps/teststeps.go
```diff
@@ -120,3 +121,38 @@ func ForEachTarget(pluginName string, cancel, pause <-chan struct{}, ch test.Tes
...
+	defer cancelFunc()
+
+	return ForEachTarget(pluginName, cancel, pause, ch, func(cancel, pause <-chan struct{}, target *target.Target) error {
+		return f(ctx, target)
+	})
+}
+
+func newTerminateContext(cancel, pause <-chan struct{}) (context.Context, context.CancelFunc) {
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	go func() {
```
rihter007, line 148: This is so bad to create an extra goroutine, I wish it could be possible to override <- in golang

rihter007:
Of the concept is ok, I will make some unit tests.

plugins/teststeps/teststeps.go
xaionaro: > so bad

Out of curiosity: why?

xaionaro (approved):
LGTM. And I like the approach of converting signal channels to a context. I was hoping ConTest will start moving to this direction :)

plugins/teststeps/teststeps.go
rihter007: Some extra overhead instead of <- overriding

rihter007:
Is obsolete, we agreed to use statectx.Context

PR #188, opened 2020-12-01
Added missing copyright header

insomniacslk:


insomniacslk:
Hopefully this won't create any problem with goose's parsing of that file

marcoguerri:
Bah, this again, sorry. Doesn't the linter run as part of the integ test run in the container? Probably we already discussed this, I forgot what the outcome of this discussion was. 

> Hopefully this won't create any problem with goose's parsing of that file

Tested locally and it doesn't interfere with the parsing, anyway the integ test run will exercise the migration so it should confirm.

insomniacslk:
> Bah, this again, sorry. Doesn't the linter run as part of the integ test run in the container? Probably we already discussed this, I forgot what the outcome of this discussion was.

The public linter ignores non-go files while the internal one doesn't. I can adjust the public linter to include sql, one sec

insomniacslk:
Updated, but I have to change the regexp too, to match the sql-style comments, one sec

insomniacslk:
> Updated, but I have to change the regexp too, to match the sql-style comments, one sec

Interestingly, it returns success, but it should fail because of the different comment style

insomniacslk:
not sure why this is happening, but sounds like a bug in `checklicense`

insomniacslk:
will look at that tomorrow

insomniacslk:
@marcoguerri if that's OK I will merge this to fix the copyright header, and file an issue for checklicenses

marcoguerri:
Sorry for the latency, yes, that's perfectly fine.

marcoguerri (approved):

PR #189, opened 2020-12-02
[plugins/cmd] Added boolean emit_stdout and emit_stderr flags

insomniacslk:
If set to true, the plugin will emit stdout and/or stderr as payloads
for the respective events.
Default is not to emit anything, because events are stored on the DB,
and command output could be large.


insomniacslk:
Tested with the modified `start-literal.yaml` that sets emit_stdout and emit_stderr to true. The relevant parts in the job status:
```
                    "Data": {
                      "EventName": "CmdStdout",
                      "Payload": {
                        "Msg": "Title=Storage2345, ToUpper=STORAGE2345\n"
                      },
                      "Target": {
                        "FQDN": "",
                        "ID": "10.10.10.123",
                        "Name": "storage2345"
                      }
                    },

...

                    "Data": {
                      "EventName": "CmdStderr",
                      "Payload": {
                        "Msg": "Title=Storage2345, ToUpper=STORAGE2345\n"
                      },
                      "Target": {
                        "FQDN": "",
                        "ID": "10.10.10.123",
                        "Name": "storage2345"
                      }
                    },

```

walterchris:
LGTM

tfg13 (approved):
GTG. Is `emit_stdout` descriptive enough so people understand it means "this will be saved into the database so maybe don't write too much"?

Also question: What happens if the payload is too long and the DB truncates it, will it end up with broken, unparsable JSON again?

insomniacslk:
> GTG. Is `emit_stdout` descriptive enough so people understand it means "this will be saved into the database so maybe don't write too much"?

I think this needs to be addressed more generally with plugin documentation. We are lagging behind on that, but I would like to tackle that soon. Until then, I think comments in code should stop the gap, I've updated the code with a comment on the struct fields.

> Also question: What happens if the payload is too long and the DB truncates it, will it end up with broken, unparsable JSON again?

I added a warning about truncation, and this should be true and documented for any event that takes arbitrary payloads. Broken parsing.. is a bigger problem. It won't break the job, but it will break the status requests

tfg13:
Ok, thats what I thought. Yeah go ahead and land it, none of the two are new problems, we need to address them in a more general way in any case.

insomniacslk:
damn, git push returned error and I didn't notice. I'll push another PR with the comment change I mentioned above

insomniacslk:
done at https://github.com/facebookincubator/contest/pull/191 , @tfg13 can you please take a look?

PR #190, opened 2020-12-03
plugins/targetmanager: Add Retries to Locking within CSVTargetManager

walterchris:
This adds retry-mechanism to the CSVTargetManager. We do shuffle and
retry locking the targets three times.

### DIFF
--- a/pkg/target/target.go
+++ b/pkg/target/target.go
@@ -56,6 +56,7 @@ func FilterTargets(targetIDs []string, targets []*Target) ([]*Target, error) {
 			if target.ID == targetID {
 				targetList = append(targetList, target)
 				found = true
+				break
 			}
 		}
 
--- a/pkg/target/target.go
+++ b/pkg/target/target.go
@@ -44,3 +44,25 @@ func (t *Target) String() string {
 	}
 	return fmt.Sprintf("Target{Name: \"%s\", ID: \"%s\", FQDN: \"%s\"}", t.Name, t.ID, t.FQDN)
 }
+
+// FilterTargets - Filter targets from targets based on targetIDs
+func FilterTargets(targetIDs []string, targets []*Target) ([]*Target, error) {
+
+	var targetList []*Target
+
+	for _, targetID := range targetIDs {
+		found := false
+		for _, target := range targets {
+			if target.ID == targetID {
+				targetList = append(targetList, target)
+				found = true
+			}
+		}
+
+		if !found {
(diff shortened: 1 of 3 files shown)


tfg13 (approved):
This should work and it is an improvement over the current version, so GTG.

In case you are interested, we had similar issues with the internal targetmanager, and I recently improved the locking API, you could do something like this now (pseudocode):

```
if (shuffle) {
  // shuffle
}
// feed all devices into new API call `TryLock`, with desired limit
locked = tl.TryLock(hosts, acquireParameters.MaxNumberOfDevices)

// check if we got enough
if len(locked) >= MinNumberOfDevices {
  // done, we got enough and they are locked
} else {
  // not enough, unlock what we got and fail
  if len(locked) > 0 {
    tl.Unlock(locked)
  }
  return fmt.Errorf("can't lock enough targets")
}
```

The advantage here is there are no probabilistic loops, it is guaranteed to lock the amount of targets you want, if possible.

plugins/targetmanagers/csvtargetmanager/csvfile.go
```diff
@@ -155,15 +156,30 @@ func (tf *CSVFileTargetManager) Acquire(jobID types.JobID, cancel <-chan struct{
 			hosts[i], hosts[j] = hosts[j], hosts[i]
 		})
 	}
-	if uint32(len(hosts)) > acquireParameters.MaxNumberDevices {
-		hosts = hosts[:acquireParameters.MaxNumberDevices]
+
+	// feed all devices into new API call `TryLock`, with desired limit
+	lockedString, err := tl.TryLock(jobID, hosts, uint(acquireParameters.MaxNumberDevices))
+	if err != nil {
+		return nil, fmt.Errorf("can't lock enough targets")
```
tfg13, line 163: The API only guarantees the upper limit, no minimum, it can return with zero targets but no error
```suggestion
		return nil, fmt.Errorf("failed to lock targets: %w", err)
```

pkg/target/target.go
```diff
@@ -44,3 +44,25 @@ func (t *Target) String() string {
...
+func FilterTargets(targetIDs []string, targets []*Target) ([]*Target, error) {
+
+	var targetList []*Target
+
+	for _, targetID := range targetIDs {
+		found := false
+		for _, target := range targets {
+			if target.ID == targetID {
+				targetList = append(targetList, target)
+				found = true
```
tfg13, line 58:
```suggestion
				found = true
				break
```

PR #193, opened 2020-12-08
[db][migration] Improve extended descriptor migration

marcoguerri:
A significant limitation of the library we use for db migrations is
that a migration can run only within a transaction. This doesn't scale
for large migrations. This patch introduces an interface to run a
migration outside of a `*sql.Tx`. The actual tool which invokes the
migration does not support running the transaction within `*sql.DB` yet:
we either need to add support to the upstream library, or move to a
custom implementation.

In addition, this patch modifies extended descriptor migration to run
batch update queries. The batch and single row implementations were
tested against ~95k records and resulted in roughly the following
numbers:
* Batch update (shard size 50): 4m26s
* Single row update (shard size 50): 5m51s

### DIFF
--- a/db/rdbms/migration/0002_migrate_descriptor_to_extended_descriptor.go
+++ b/db/rdbms/migration/0002_migrate_descriptor_to_extended_descriptor.go
@@ -9,10 +9,12 @@ import (
 	"database/sql"
 	"encoding/json"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/facebookincubator/contest/pkg/job"
 	"github.com/facebookincubator/contest/pkg/test"
+	"github.com/facebookincubator/contest/pkg/types"
 	"github.com/facebookincubator/contest/tools/migration/rdbms/migrate"
 
 	"github.com/facebookincubator/contest/cmds/plugins"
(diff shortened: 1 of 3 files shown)


db/rdbms/migration/0002_migrate_descriptor_to_extended_descriptor.go
```diff
@@ -58,21 +60,32 @@ const shardSize = uint64(50)
 // jobs table.
 type DescriptorMigration struct {
 	log *logrus.Entry
+	db  *sql.DB
+}
+
+type dbConn interface {
+	Exec(query string, args ...interface{}) (sql.Result, error)
+	Query(query string, args ...interface{}) (*sql.Rows, error)
+}
```
marcoguerri, line 69: I will probably move this somewhere else. It is actually an interface implemented also in the rdbms plugin, but I'd rather duplicate, than introduce a dependency between the two.

marcoguerri:
Linters of course are failing, my new env is misconfigured and it's not actually running the linters I would like it to run. Trying to fix it :)

db/rdbms/migration/0002_migrate_descriptor_to_extended_descriptor.go
```diff
@@ -60,19 +62,29 @@ type DescriptorMigration struct {
...
 
+type dbConn interface {
+	Exec(query string, args ...interface{}) (sql.Result, error)
+	Query(query string, args ...interface{}) (*sql.Rows, error)
+}
+
+type jobDescriptorPairs struct {
+	jobIDs              []types.JobID
+	extendedDescriptors []string
+}
```
tfg13, line 73: Just a nit, why not make this
```
type jobDescriptorPair struct {
	jobID              types.JobID
	extendedDescriptor string
}
```
and then below having
```
var updates []jobDescriptorPair
```
so they are guaranteed to stay together without being super careful with the indices?

tfg13 (approved):
Just a nit, you don't need to refactor it again

db/rdbms/migration/0002_migrate_descriptor_to_extended_descriptor.go
marcoguerri: Heh, I have gone through several iterations, and your proposal was definitely one of the approaches, in one of the intermediate states. I don't think there is any reason not to adopt that approach in this version, which I like much more.

PR #194, opened 2020-12-08
[storage] Enforce the desired db version

marcoguerri:
We started tracking the db schema version. The code will have a
hard dependency on a specific schema version being available, so
we should fail any attempt to set a storage backend if its advertised
schema version does not match the requirements of core ConTest storage
pkg..

For all those instances for which schema versioning is not in place yet
(e.g. schema is v0, so the reference baseline), reported version will
be 0. This patch sets the minimum schema version to 0, so it will allow
any backend version.

### DIFF
--- a/tools/migration/rdbms/migrationlib/version.go
+++ b/tools/migration/rdbms/migrationlib/version.go
@@ -6,11 +6,78 @@
 package migrationlib
 
 import (
+	"fmt"
+
 	"database/sql"
+	"github.com/go-sql-driver/mysql"
 	"github.com/pressly/goose"
 )
 
-// DBVersion returns the current version of the database schema
-func DBVersion(db *sql.DB) (int64, error) {
-	return goose.GetDBVersion(db)
+// queryExecutor represents the MCD interface between sql.Db and sql.Tx
+// objects that is sufficient to retrieve the db version number.
+type queryExecutor interface {
+	Query(query string, args ...interface{}) (*sql.Rows, error)
+}
+
+// DBVersion returns the current version of the database schema.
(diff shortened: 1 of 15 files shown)


tools/migration/rdbms/migrationlib/version.go
```diff
@@ -6,11 +6,68 @@
...
+type queryExecutor interface {
+	Query(query string, args ...interface{}) (*sql.Rows, error)
+}
+
+// DBVersion returns the current version of the database schema.
+// The library exposes a GetDBVersion function which however requires
+// a *sql.Db object. Several contest flows work instead within a
+// transaction al context, so we need to be able to extract the version
+// in a way that is agnostic to the library.
+func DBVersion(db queryExecutor) (uint64, error) {
```
marcoguerri, line 26: I could not find a way to avoid this, unfortunately. I am checking with upstream library maintainer if they accept PRs to make the library more flexible in terms of what it supports (`sql.Tx` vs `sql.Db` storage). Until then, we need stay on this stopgap. If nothing moves upstream I think we might want to consider building our own small library.

tfg13 (approved):
LGTM.

OneQ: How does it react if the version table does not exist (yet)?
The `DBVersion ` stuff will error out as far as I can tell (same if it exists but is empty) - is this caught somewhere else in goose?

marcoguerri:
Right, good point. The initial version of this PR was using the library, which does take care of the non-versioned instances. But then I stumbled across the issue with `sql.Tx `, which we do use in several test flows and re-implemented the version check, and lost the safety for non-versioned instances. Thank you for spotting this, I need to fix this.

PR #196, opened 2020-12-09
State context implementation and usage in job.Job

rihter007:
Implement types.StateContext that allows to pause/cancel this context returning different error codes for each case.

Use StateContext in job.Job instead of cancel/pause channels.
Use types.StateContext in plugins.

Next: introduce types.StateContext for teststeps


pkg/runner/test_runner.go
```diff
@@ -197,19 +198,11 @@ func (tr *TestRunner) Run(cancel, pause <-chan struct{}, test *test.Test, target
 	// the pipeline.
 	for {
 		select {
-		case <-cancel:
```
rihter007, line 200: We do not need to wait for cancel/pause. If this event occures, testPipeline.run is expected to terminate and will write something to errCh

pkg/statectx/cancel_context.go
```diff
@@ -0,0 +1,57 @@
+// Note: unfortunately context.WithCancel can't redefine error code
```
rihter007, line 1: It is such a shame that context.WithCancel implementation is private

pkg/statectx/context.go
```diff
@@ -0,0 +1,91 @@
...
+var ErrCanceled = errors.New("job is canceled")
+
+// ErrPaused is returned by Context.Err when the context was paused
+var ErrPaused = errors.New("job is paused")
+
+// statectx.Context implements context.Context interface that acts as context for cancellation signal
+// It also implements:
+// - OnPause()/PauseCtx() that alert in case of pause happened
+// - Any()/AnyCtx() that alert in case of pause/cancel events
+// - State()
```
rihter007, line 24: State is no longer needed, should remove this comment

pkg/statectx/context.go
```diff
@@ -0,0 +1,91 @@
...
+type Context interface {
```
rojer, line 25: i think it's ok but naming is somewhat inconsistent, we have:
```
	case <-ctx.Done():
	case <-ctx.Any():
	case <-ctx.OnPause():
```

i think this would be more consistent and readable:

```
	case <-ctx.Done():
	case <-ctx.Paused(): // instead of OnPause
	case <-ctx.PausedOrDone():  // instead of Any
```

PR #199, opened 2020-12-10
[db][migration] Add initial documentation for db migration tool

marcoguerri:
This documentation covers the tool only. Additional documentation should follow for the migrations themselves.


tools/migration/rdbms/README.md
```diff
@@ -0,0 +1,99 @@
...
+2020/12/09 19:02:45     Applied At                  Migration
+2020/12/09 19:02:45     =======================================
+2020/12/09 19:02:45     Pending                  -- 0001_add_extended_descriptor.sql
+2020/12/09 19:02:45     Pending                  -- 0002_migrate_descriptor_to_extended_descriptor.go
+```
+
+We can now proceed applying the migrations.
+
+# Up command
+`up` command, and its variations, can be used run forward migrations. Taking for example the output of the `status` command above, we can run both migrations in sequence:
```
tfg13, line 68:
```suggestion
`up` command, and its variations, can be used to run forward migrations. Taking for example the output of the `status` command above, we can run both migrations in sequence:
```

tools/migration/rdbms/README.md
```diff
@@ -0,0 +1,99 @@
...
+```
+ $ go run tools/migration/rdbms/main.go -dir db/rdbms/migration status
+2020/12/09 19:12:05     Applied At                  Migration
+2020/12/09 19:12:05     =======================================
+2020/12/09 19:12:05     Wed Dec  9 19:11:04 2020 -- 0001_add_extended_descriptor.sql
+2020/12/09 19:12:05     Wed Dec  9 19:11:04 2020 -- 0002_migrate_descriptor_to_extended_descriptor.go
+```
+
+# Down command
+`down` is the counterpart of `up`, as it rewinds the migration. Note that what actually happens when a migration is executed on the `down` path is very dependent on the implementation of the migration itself. `down` migrations might be a no-op, in which case the tool will only roll back the version, but no actual work will be executed.
```
tfg13, line 99: Is the no-op different from a down migration not being supported, or is that the same thing?

If they are the same, should we not fail and say "Sry can't go down"?

tfg13 (approved):

tools/migration/rdbms/README.md
marcoguerri: It's the same, and I have been thinking of making it fail hard, but I couldn't find a non-ugly way to do so in `.sql` migrations and during testing it was convenient to be able to reset the db version. But you are right, failing hard is the right thing to do. 

I'll sort out the first internal migration and put up a PR to do so.

PR #203, opened 2020-12-11
Use state context in test steps plugins

rihter007:
Use statectx.Context in teststeps plugins


pkg/runner/test_runner_pipeline.go
```diff
@@ -169,15 +162,12 @@ func (p *pipeline) runStep(cancel, pause <-chan struct{}, jobID types.JobID, run
 		}
 		// stepCh.stepIn is not closed, but the TestStep returned, which is a violation
 		// of the API. Record the error if no other error condition has been seen.
-		if err == nil {
```
rihter007, line 172: above if err != nil -> here err == nil is always trye

rojer:
this looks reasonable to me but please wait for a real review :)

rihter007:
> this looks reasonable to me but please wait for a real review :)

I see that you have implemented some similar changes as well + supported resume semantics in the new runner.
It might be the case that this PR is not needed anymore.

rojer:
no, we still need to convert plugins to use context, so this one should go ahead

pkg/runner/test_runner_route.go
```diff
@@ -58,7 +50,8 @@ func (r *stepRouter) routeIn(terminate <-chan struct{}) (int, error) {
 
 	// terminateTargetWriter is a control channel used to signal termination to
 	// the writer object which injects a target into the test step
-	terminateTargetWriter := make(chan struct{})
+	terminateTargetWriterCtx, terminateTargetWriter := context.WithCancel(context.Background())
+	defer terminateTargetWriter() // avoids possible goroutine deadlock in context.WithCancel implementation
```
tfg13, line 54: Why do we need this, can this ever fire before we hit the `terminateTargetWriter()` below?

plugins/teststeps/teststeps_test.go
```diff
@@ -6,7 +6,9 @@
 package teststeps
 
 import (
+	"context"
 	"fmt"
+	"github.com/facebookincubator/contest/pkg/statectx"
```
tfg13, line 11: move to other imports further below (also check the other files, just noted it here)

plugins/teststeps/teststeps_test.go
```diff
@@ -217,12 +225,12 @@ func TestForEachTargetTenTargetsParallelism(t *testing.T) {
 	logging.Debug()
 	sleepTime := time.Second
 	d := newData()
-	fn := func(cancel, pause <-chan struct{}, tgt *target.Target) error {
+	fn := func(ctx statectx.Context, tgt *target.Target) error {
 		log.Printf("Handling target %+v", tgt)
 		select {
-		case <-cancel:
+		case <-ctx.Done():
 			log.Printf("target %+v caneled", tgt)
```
tfg13, line 232: Since we are here already (also in more comments in this file)
```suggestion
			log.Printf("target %+v cannelled", tgt)
```

tfg13:
Mostly LGTM, some nits and one question

pkg/runner/test_runner_route.go
rihter007: This change is not required, I just decided to get rid off channels.

This can't fire before we reach terminateTargetWriter() because we create a channel around context.Background() (not on the input ctx) which never gets cancelled.

So, it should be the same logic as it was before, but just using context.

tfg13: What I meant was, why do we need `defer terminateTargetWriter()` here, when a couple lines later we unconditionally call the same function anyway (terminateTargetWriter()). This is before the defer would fire.

tfg13 (approved):

PR #205, opened 2020-12-15
Implement context inheritance and cancel/pause signals propagation for statectx.Context

rihter007:
Runner is required to make statectx.Context that it provides to its teststeps and it needs to propagate cancel/pause signals that it takes as a statectx.Context from contest

### DIFF
--- a/pkg/statectx/cancel_context.go
+++ b/pkg/statectx/cancel_context.go
@@ -99,14 +99,6 @@ func (c *cancelContext) propagateCancel(parent context.Context) {
 		return // parent is never canceled
 	}
 
-	select {
-	case <-done:
-		// parent is already canceled
-		c.cancel(parent.Err())
-		return
-	default:
-	}
-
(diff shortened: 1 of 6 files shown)


pkg/statectx/cancel_context.go
```diff
@@ -8,10 +8,11 @@ import (
 	"time"
 )
 
-func newCancelContext() (context.Context, func(err error)) {
+func newCancelContext(ctx context.Context) (context.Context, func(err error)) {
 	result := &cancelContext{
 		done: make(chan struct{}),
 	}
+	result.propagateCancel(ctx)
```
rihter007, line 15: Almost copy-paste of context.WithCancel that unfortunately doesn't support custom error codes

rojer:
i see a bunch of missing coverage warnings that seem relevant.
this is tricky code, can you ensure that all the cases are covered by unit tests?

rojer (approved):

PR #122, opened 2020-06-16
[test_runner] Unify wait routines

marcoguerri:
The pipeline code was affected by some duplication in the shutdown
control path. This patch removes the duplication, and introduces
only two wait routines in the shutdown control path: waitTargets
and waitControl.

PR #207, opened 2020-12-16
Minor refactor stepRouter routeIin

rihter007:


rihter007:
Why do we have that injection logic?
Why not to return results as in a normal function?

pkg/runner/test_runner_route.go
```diff
@@ -45,109 +43,40 @@ func tryClose(ch chan struct{}) {
...
-		// no targets currently being injected in the test step
-		if targets.Len() == 0 {
-			if r.routingChannels.routeIn == nil {
-				log.Debugf("input channel is closed and no more targets are available, closing step input channel")
 				close(r.routingChannels.stepIn)
-				break
+				return len(ingressTarget), nil
+			}
+			log.Debugf("received target %v in input", t)
+			ingressTarget[t.ID] = time.Now()
```
rihter007, line 65: Is there any chance that targets will be repeated?
If not it could be just targetsCount++

rihter007:
Talked in person. Seems to be useless because @rojer redeveloped the whole concept of test runner.

Nevertheless @marcoguerri mentioned that one of the goals of this code was to accumulate all targets from the input channels. That logic is missing here. (In my view this is the only thing that changed, thanks @marcoguerri for explaining the logic behind that code)

marcoguerri:
Yes, sorry about the communication gap. I had several PRs in the queue to simplify the current runner (#119, #121, #122) and in the last one I actually took it a stab at getting rid at all of that targets buffer, which is excessively complex (in particular, inhttps://github.com/facebookincubator/contest/pull/122/commits/c53e4269ef940b51fd6eed7d85dbf5656fe84000). 

The idea of that buffer is that as soon as a target it returned by the test step, you want to take it in charge, and not block the step. I reworked it in https://github.com/facebookincubator/contest/pull/122/commits/c53e4269ef940b51fd6eed7d85dbf5656fe84000 so that you acquire it, assign it to a goroutine, and then make them all forward to the next routing block. If the next routing block encounters a timeout in injection, it will generate a termination of the whole block, so the pending goroutines will be terminated as well.

But as I mentioned, while working on resume, a re-worked test runner is being implemented, so we shouldn't invest in additional changes to the current test runner. Sorry for the communication gap.

PR #208, opened 2020-12-17
New test runner implementation

rojer:
Simplified, added support for serializing state and resuming (not used yet but will be).


rojer:
> will **decrease** coverage by `1.66%`.

new runner is 91% covered by tests but since more code is deleted than added, overall coverage goes down.

pkg/runner/test_runner.go
```diff
@@ -1,287 +1,750 @@
...
+		return tr.checkStepRunners()
+	case <-time.After(tr.shutdownTimeout):
+		tr.log.Errorf("step runners failed to shut down correctly")
+		// If there is a step with an error set, use that.
+		if err := tr.checkStepRunners(); err != nil {
+			return err
+		}
+		// If there isn't, enumerate ones that were still running at the time.
+		err := &cerrors.ErrTestStepsNeverReturned{}
+		tr.mu.Lock()
```
rihter007, line 358: Because we acquire the lock in the goroutine above, we get blocked here until that goroutine finishes (swch channel is closed)

rojer: not quite. cond.Wait() releases the lock while waiting.

rihter007: Sorry, it seems to be ok because of tr.cond.Wait() will release the mutex

rojer:
clarified the description a bit, on feedback from @rihter007

marcoguerri:
I haven't started full review yet, it will take a while. What this is missing is emissions of `TargetIn` (https://github.com/facebookincubator/contest/blob/master/pkg/runner/test_runner_route.go#L88) and `TargetOut` events (https://github.com/facebookincubator/contest/blob/master/pkg/runner/test_runner_route.go#L204), which are then used to rebuild the status of the job (https://github.com/facebookincubator/contest/blob/master/pkg/runner/job_status.go#L23-L28) from any instance, even those that do not own the job.

This is duplication of how we keep track of the state and we should converge into using the same approach both for resume and and Status queries.

rojer:
@marcoguerri thanks! indeed, i did not pay sufficient attention to the events that should be emitted. this should be fixed now - i added TargetIn, TargetOut, TargetErr and made sure a TestError is emitted in the right cases.

pkg/runner/test_runner.go
```diff
@@ -6,282 +6,778 @@
...
-func newTargetWriter(log *logrus.Entry, timeouts TestRunnerTimeouts) *targetWriter {
-	return &targetWriter{log: log, timeouts: timeouts}
+// targetRunner runs one target through all the steps of the pipeline.
+func (tr *testRunner) targetRunner(ctx, stepCtx statectx.Context, ts *targetState) {
+	log := logging.AddField(tr.log, "target", ts.tgt.ID)
+	log.Debugf("%s: target runner active", ts)
+	// NB: CurStep may be non-zero on entry if resumed
+loop:
+	for i := ts.CurStep; i < len(tr.steps); {
+		// Early check for pause of cancelation.
```
tfg13, line 377:
```suggestion
		// Early check for pause or cancelation.
```

pkg/runner/test_runner.go
```diff
@@ -6,282 +6,778 @@
...
+	for {
+		select {
+		case tgt, ok := <-outCh:
+			if !ok {
+				ss.log.Debugf("%s: out chan closed", ss)
+				// At this point we may still have an error to report,
+				// wait until error channel is emptied too.
+				outCh = make(chan *target.Target)
+				tr.safeCloseErrCh(ss)
+				break
```
tfg13, line 625:
```suggestion
				break loop
```

pkg/runner/test_runner.go
```diff
@@ -6,282 +6,778 @@
...
+// and closes input channels of the steps to indicate that no more are expected.
+// It also monitors steps for critical errors and cancels the whole run.
+// Note: input channels remain open when cancellation is requested,
+// plugins are expected to handle it explicitly.
+func (tr *testRunner) runMonitor() error {
+	tr.log.Debugf("monitor: active")
+	tr.mu.Lock()
+	defer tr.mu.Unlock()
+	// First, compute the starting step of the pipeline (it may be non-zero
+	// if the pipleine was resumed).
```
tfg13, line 690:
```suggestion
	// if the pipeline was resumed).
```

tfg13 (commented):
Mostly nits. Since it is a super large change, it would be good if someone else would also read it as well.

pkg/runner/test_runner.go
rojer9-fb: actually, on further reflection - no, we don't want to break out of the loop just yet. as the comment above says, we want to make sure error channel also gets processed, which is exactly why we replace outCh with a new one and wait for ss.errCh to be closed as well.

pkg/runner/goroutine_leak_check.go
```diff
@@ -0,0 +1,103 @@
...
+	Line  int
+	Func  string
+	Trace []string
+}
+
+func (e *stackTraceEntry) String() string {
+	return fmt.Sprintf("%s:%d %s %d", e.File, e.Line, e.Func, e.GoID)
+}
+
+func CheckLeakedGoRoutines(funcWhitelist []string) error {
```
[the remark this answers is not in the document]
rojer: done

pkg/jobmanager/jobmanager.go
```diff
@@ -158,7 +158,7 @@ func newPartialJobFromDescriptor(pr *pluginregistry.PluginRegistry, jd *job.JobD
 			}
 			// test step index is incremented by 1 so we can use 0 to signal an
 			// anomaly.
```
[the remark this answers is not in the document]
rojer: gone

pkg/runner/test_runner.go
tfg13: Ah wasn't sure what you want to break out of, I guess you meant the select then?
Would `continue loop` be clearer in that case?

rojer: great idea, indeed that would make the intent more clear. done.

pkg/runner/test_runner.go
```diff
@@ -6,282 +6,778 @@
...
+			break loop
+		}
+		// Await result. It will be communicated to us by the step runner.
+		select {
+		case res, ok := <-ts.resCh:
+			if !ok {
+				log.Debugf("%s: result channel closed", ts)
+				break loop
+			}
+			log.Debugf("%s: result for %s recd", ts, ss)
```
xaionaro, line 440: What is `recd`?

pkg/runner/test_runner.go
```diff
@@ -6,282 +6,778 @@
...
-	stepOut <-chan *target.Target
-	stepErr <-chan cerrors.TargetError
-	// targetErr connects the routing block directly to the TestRunner. Failing
-	// targets are acquired by the TestRunner via this channel
-	targetErr chan<- cerrors.TargetError
+// testRunner is the state associated with a test run.
+// Here's how a test run works:
+//  * Each target gets a targetState and a "target runner" - a goroutine that takes that particular
+//    target through each step of the pipeline in sequence. It injects the target, waits for the result,
+//    then mves on to the next step.
```
xaionaro, line 41: mves -> moves

pkg/runner/test_runner.go
```diff
@@ -6,282 +6,778 @@
...
-	return r.completedRouting
+// runMonitor monitors progress of targets through the pipeline
+// and closes input channels of the steps to indicate that no more are expected.
+// It also monitors steps for critical errors and cancels the whole run.
+// Note: input channels remain open when cancellation is requested,
+// plugins are expected to handle it explicitly.
+func (tr *testRunner) runMonitor() error {
+	tr.log.Debugf("monitor: active")
+	tr.mu.Lock()
+	defer tr.mu.Unlock()
```
xaionaro, line 688: I see you lock the mutex for the whole runMonitor execution. How `targetRunners` works in this case?

pkg/runner/test_runner.go
```diff
@@ -6,282 +6,778 @@
...
-// writeTargetError writes a TargetError object to a TargetError channel with timeout
-func (w *targetWriter) writeTargetError(ctx context.Context, ch chan<- cerrors.TargetError, targetError cerrors.TargetError, timeout time.Duration) error {
+func (tr *testRunner) waitStepRunners(ctx statectx.Context) error {
+	tr.log.Debugf("waiting for step runners to finish")
+	swch := make(chan struct{})
+	go func() {
+		tr.mu.Lock()
+		defer tr.mu.Unlock()
+		for {
+			ok := true
```
xaionaro, line 303: Could you rename this variable to something more meaningful, please?

pkg/runner/test_runner.go
```diff
@@ -6,282 +6,778 @@
...
+			}
+		}
 	}
+
+	return resumeState, runErr
 }
 
-// writeTargetError writes a TargetError object to a TargetError channel with timeout
-func (w *targetWriter) writeTargetError(ctx context.Context, ch chan<- cerrors.TargetError, targetError cerrors.TargetError, timeout time.Duration) error {
+func (tr *testRunner) waitStepRunners(ctx statectx.Context) error {
```
xaionaro, line 296: `ctx` is not used.

pkg/runner/test_runner.go
```diff
@@ -6,282 +6,778 @@
...
+		if runErr == nil {
+			runErr = err
+		}
+	}
+	// There will be no more results, reel in all the target runners (if any).
+	tr.log.Debugf("waiting for target runners to finish")
+	targetCancel()
+	tr.targetsWg.Wait()
+
+	// Examine the resulting state.
```
xaionaro, line 229: Could you please elaborate in this comment?

pkg/runner/test_runner.go
```diff
@@ -6,282 +6,778 @@
...
+			if stepErr != statectx.ErrPaused {
+				resumeOk = false
+			}
+		}
+	}
+	tr.log.Debugf("- %d in flight, ok to resume? %t", len(inFlightTargets), resumeOk)
+	tr.log.Debugf("step states:")
+	for i, ss := range tr.steps {
+		tr.log.Debugf("  %d %s %t %t %v", i, ss, ss.stepRunning, ss.readerRunning, ss.runErr)
+	}
```
xaionaro, line 250: Better to use structured logs. See `WithField`/`WithFields`. It allows to perform smarter log-based analytics afterwards.

pkg/runner/test_runner.go
```diff
@@ -6,282 +6,778 @@
...
-		log := logging.AddField(log, "step", "injection")
-		writer := newTargetWriter(log, tr.timeouts)
-		for _, t := range targets {
-			if err := writer.writeTimeout(ctx, inputChannel, t, tr.timeouts.MessageTimeout); err != nil {
-				log.Debugf("could not inject target %+v into first routing block: %+v", t, err)
+// stepRunner runs a test pipeline's step (the Run() method).
+func (tr *testRunner) stepRunner(ctx statectx.Context, ss *stepState) {
+	ss.log.Debugf("%s: step runner active", ss)
+	defer func() {
+		if r := recover(); r != nil {
```
xaionaro, line 521: Why this recover is required? Could you leave a comment, please?

pkg/runner/test_runner.go
```diff
@@ -6,282 +6,778 @@
...
-	// the pipeline.
-	for {
-		select {
-		case err := <-errCh:
-			log.Debugf("test runner terminated, returning %v", err)
-			return err
-		case t := <-completedTargets:
-			log.Infof("test runner completed target: %v", t)
-		}
+	chans := test.TestStepChannels{In: ss.inCh, Out: ss.outCh, Err: ss.errCh}
```
xaionaro, line 532: Out of curiosity, why we do not initialize `test.TestStepChannels` right inside `stepState` (and instead we define three separate channels)?

pkg/runner/test_runner.go
```diff
@@ -6,282 +6,778 @@
...
+	for {
+		select {
+		case tgt, ok := <-outCh:
+			if !ok {
+				ss.log.Debugf("%s: out chan closed", ss)
+				// At this point we may still have an error to report,
+				// wait until error channel is emptied too.
+				outCh = make(chan *target.Target)
+				tr.safeCloseErrCh(ss)
+				continue loop
```
xaionaro, line 625: `loop` is not required here

pkg/runner/test_runner.go
```diff
@@ -6,282 +6,778 @@
...
+			}
+			if err = tr.reportTargetResult(ctx, ss, tgt, nil); err != nil {
+				break loop
+			}
+		case res, ok := <-ss.errCh:
+			if !ok {
+				ss.log.Debugf("%s: err chan closed", ss)
+				break loop
+			}
```
xaionaro, line 634: It looks racy for me:
What will happen if `outCh` is closed, but `errCh` contains an error? Or this is intended?

pkg/runner/test_runner.go
```diff
@@ -6,282 +6,778 @@
...
+	outCh := ss.outCh
+loop:
+	for {
+		select {
+		case tgt, ok := <-outCh:
+			if !ok {
+				ss.log.Debugf("%s: out chan closed", ss)
+				// At this point we may still have an error to report,
+				// wait until error channel is emptied too.
+				outCh = make(chan *target.Target)
```
xaionaro, line 623: Why not `outCh = nil`?

pkg/runner/test_runner.go
```diff
@@ -6,282 +6,778 @@
...
+				break loop
+			}
+		case <-ctx.Done():
+			ss.log.Debugf("%s: canceled 3, draining", ss)
+			for {
+				select {
+				case <-outCh:
+					break loop
+				case <-ss.errCh:
+					break loop
```
xaionaro, line 645: It does not look like proper draining: I guess there could be multiple events in `ss.errCh`.

pkg/runner/test_runner.go
```diff
@@ -6,282 +6,778 @@
...
+				case <-time.After(tr.shutdownTimeout):
+					tr.mu.Lock()
+					if ss.runErr == nil {
+						ss.runErr = &cerrors.ErrTestStepsNeverReturned{}
+					}
+					tr.mu.Unlock()
```
xaionaro, line 651: So we timed-out but not interrupted the loop? Is this correct?

xaionaro (approved):

xaionaro:
Just a warning: the quality of my review was poor this time.

pkg/cerrors/cerrors.go
```diff
@@ -53,3 +64,47 @@ type ErrTestStepClosedChannels struct {
 func (e *ErrTestStepClosedChannels) Error() string {
 	return fmt.Sprintf("test step %v closed output channels (api violation)", e.StepName)
 }
+
+// ErrTestStepPaniced indicates that a test step's method paniced.
+type ErrTestStepPaniced struct {
```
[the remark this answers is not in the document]
rojer: done

pkg/runner/job_runner.go
```diff
@@ -183,7 +183,14 @@ func (jr *JobRunner) Run(j *job.Job) ([][]*job.Report, []*job.Report, error) {
 			if runErr = jr.emitAcquiredTargets(testEventEmitter, targets); runErr == nil {
 				jobLog.Infof("Run #%d: running test #%d for job '%s' (job ID: %d) on %d targets", run+1, idx, j.Name, j.ID, len(targets))
 				testRunner := NewTestRunner()
-				runErr = testRunner.Run(j.StateCtx, t, targets, j.ID, types.RunID(run+1))
+				var resumeState []byte
+				resumeState, err := testRunner.Run(j.StateCtx, t, targets, j.ID, types.RunID(run+1), resumeState)
+				if err == statectx.ErrPaused {
+					jobLog.Debugf("Runner paused, state: %s", string(resumeState))
+					// TODO(rojer): Persist the state.
```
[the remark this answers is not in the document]
rojer: done

pkg/runner/test_runner.go
```diff
@@ -6,282 +6,778 @@
...
 )
 
-// TestRunnerTimeouts collects all the timeouts values that the test runner uses
-type TestRunnerTimeouts struct {
-	StepInjectTimeout   time.Duration
-	MessageTimeout      time.Duration
-	ShutdownTimeout     time.Duration
-	StepShutdownTimeout time.Duration
+// TestRunner is the interface test runner implements
+type TestRunner interface {
```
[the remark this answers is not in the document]
rojer: done

pkg/runner/test_runner.go
rojer: it's not but i think it's more readable that way

pkg/runner/test_runner.go
rojer: indeed. i forgot that reading from nil channel blocks.

pkg/runner/test_runner.go
rojer: out and err are unbuffered and are closed only after step's run function has exited or if we gave up on waiting for it to exit.
step is not allowed to leave its run function without successfully processing targets so we don't need to care about some targets getting stuck. if the step left its run method, it means all targets have been processed.

pkg/runner/test_runner.go
rojer: we are interrupting the loop - just below this we forcibly close the output channel even though the run method is still running precisely to release the reader. the reader will close error channel and exit, and it will also release the goroutine we started above. as a result, the only leak will be of the goroutine stuck in the step's Run method (this is verified in the test using CheckLeakedGoRoutines)

xaionaro:
P.S.: Also mutex-logic looks dangerous in multiple ways. I guess it might cause a deadlock somewhere in future (when somebody will modify the code).

pkg/runner/goroutine_leak_check.go
```diff
@@ -0,0 +1,103 @@
...
+	buf := make([]byte, 1000000)
+	n := runtime.Stack(buf, true /* all */)
+	// First goroutine is always the running one, we skip it.
+	ph := 0
+	var e *stackTraceEntry
+	var badEntries []string
+	for _, line := range strings.Split(string(buf[:n]), "\n") {
+		switch ph {
+		case 0: // Look for an empty line
+			if len(line) == 0 {
```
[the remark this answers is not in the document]
rojer: added a test

pkg/runner/test_runner.go
rojer: that's precisely the expected sequence: out is closed, then error. when out is closed, we continue to wait for error to be closed, as explained in the comment above.

pkg/runner/test_runner.go
```diff
@@ -6,282 +6,778 @@
...
-	pause  func()
-	cancel func()
+// targetStepPhase denotes progression of a target through a step
+type targetStepPhase int
+
+const (
+	targetStepPhaseInit  targetStepPhase = 0
+	targetStepPhaseBegin targetStepPhase = 1 // Picked up for execution.
+	targetStepPhaseRun   targetStepPhase = 2 // Injected into step.
+	targetStepPhaseEnd   targetStepPhase = 3 // Finished running a step.
```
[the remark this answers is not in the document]
rojer: ok, done

rojer:
> P.S.: Also mutex-logic looks dangerous in multiple ways. I guess it might cause a deadlock somewhere in future (when somebody will modify the code).

well, this is overly general. previous implementation didn't use mutexes but was a maze of channels that was impossible to understand.
in this refactoring i tried to keep locking simple. in particular by having only one mutex, even though using separate per-target and per-step locks might've been more efficient but would make code considerably more complicated.

pkg/runner/test_runner.go
```diff
@@ -6,282 +6,796 @@
...
+			return nil, fmt.Errorf("wrong resume state, job id %d (want %d)", rs.JobID, jobID)
+		}
+		tr.targets = rs.Targets
+	}
+	// Initialize remaining fields of the target structures,
+	// build the map and kick off target processing.
+	for _, tgt := range targets {
+		ts := tr.targets[tgt.ID]
+		if ts == nil {
+			ts = &targetState{
```
marcoguerri, line 191: This new `targetState` doesn't get backported in `tr.targets`?

EDIT: nevermind, it happens down below.

pkg/runner/test_runner.go
```diff
@@ -6,282 +6,796 @@
...
+	runErr := tr.runMonitor()
+	if runErr != nil {
+		tr.log.Errorf("monitor returned error: %q, canceling", runErr)
+		stepCancel()
+	}
+
+	// Wait for step runners and readers to exit.
+	if err := tr.waitStepRunners(ctx); err != nil {
+		tr.log.Errorf("step runner error: %q, canceling", err)
+		stepCancel()
```
marcoguerri, line 217: Does `waitStepRunners` guarantee that all steps have returned? If so, `stepCancel` should not be necessary. If not, steps should be awaited for if a cancellation signal is being propagated. We should keep track of any step that does not return within the timeout and flag it accordingly.

pkg/runner/test_runner.go
```diff
@@ -6,282 +6,796 @@
...
+	defer tr.mu.Unlock()
+	resumeOk := (runErr == nil)
+	var inFlightTargets []*targetState
+	for i, tgt := range targets {
+		ts := tr.targets[tgt.ID]
+		stepErr := tr.steps[ts.CurStep].runErr
+		tr.log.Debugf("  %d %s %v", i, ts, stepErr)
+		if ts.CurPhase == targetStepPhaseRun {
+			inFlightTargets = append(inFlightTargets, ts)
+			if stepErr != statectx.ErrPaused {
```
marcoguerri, line 239: If there is even only one target that failed in any step, we are declaring the test not resumable (but continuing the loop after we declare `resumeOk` false)? It should be fine to fail a target in a step, and still be able to resume the test.

pkg/runner/test_runner.go
```diff
@@ -6,282 +6,796 @@
...
+			Targets: tr.targets,
+		}
+		resumeState, runErr = json.Marshal(rs)
+		if runErr != nil {
+			tr.log.Errorf("unable to serialize the state: %s", runErr)
+		} else {
+			runErr = statectx.ErrPaused
+		}
+	default:
+		// We are not pausing and yet some targets were left in flight.
```
marcoguerri, line 281: Ho do we get here? `runMonitor` has returned, `waitStepRunners` has returned, so I assume all steps have  returned (without an error? otherwise `runErr` would have been not nil and we would have returned before?) How do we get into this state, where we consider steps terminated, monitor terminated, but still targets might be in-flight?

marcoguerri (commented):
Apologies for the late review. In fact, I am catching up only now, and I am also far from being done (for my reference, I shall resume from `waitStepRunners`). Even if this is pushed, I still want to finish leaving a first pass of comments today.

marcoguerri:
> but was a maze of channels that was impossible to understand.

There was a lot of room for simplification in the previous implementation, in particular:
* Not using context made several things overly complicated
* All results, were waited  by the test runner. There was a lot of room for simplification to give distribute responsibility to routing blocks, and simplify test runner (gave it a shot in https://github.com/facebookincubator/contest/pull/122/commits/c53e4269ef940b51fd6eed7d85dbf5656fe84000)
* Running pipelines back to back (test + cleanup) also added quite a bit of complexity, that could be removed (commit above does this as well). 

This just to say that in my opinion this solution has its own traps, as it makes data flow implicit (via mutex and global structure) rather than explicit, via channels. No questioning that the previous solution could be simplified, of course, but I personally find this solution not significantly easier to follow, with respect to the previous version with some unnecessary complexity removed.

xaionaro:
> This just to say that in my opinion this solution has its own traps, as it makes data flow implicit (via mutex and global structure) rather than explicit, via channels. No questioning that the previous solution could be simplified, of course, but I personally find this solution not significantly easier to follow, with respect to the previous version with some unnecessary complexity removed.

I have exactly the same impression.

pkg/runner/test_runner.go
rojer: you are right, this is not needed as by this point all the step runners have either exited or we gave up waiting on them to exit.

pkg/runner/test_runner.go
rojer: stepState.runErr aka stepErr is not about target errors, it is

// Runner error, returned from Run() or an error condition detected by the reader.

i.e. it's an error condition that aborts entire run. presence of such errors making run non-resumable makes sense.

pkg/runner/test_runner.go
rojer: that's if plugin "swallowed" a target - it returned from Run with no error but never reported result for a target that was injected.

rojer:
@marcoguerri sorry, i should have waited for your approval before merging. but i created https://github.com/facebookincubator/contest/pull/212 so we can continue the review there.

pkg/runner/test_runner.go
marcoguerri: Well, ok, not sure why I missed that here we were referring to an error of the whole step.

pkg/runner/test_runner.go
```diff
@@ -6,282 +6,796 @@
...
+		break
 	}
+	return nil
 }
 
-// NewTestRunnerWithTimeouts initializes and returns a new TestRunner object with
-// custom timeouts
-func NewTestRunnerWithTimeouts(timeouts TestRunnerTimeouts) TestRunner {
-	return TestRunner{timeouts: timeouts}
+func (tr *TestRunner) safeCloseOutCh(ss *stepState) {
```
marcoguerri, line 603: Looks like `safeCloseOutCh` and `safeCloseErrCh` can be merged together and accept an argument from outside.

pkg/runner/test_runner.go
```diff
@@ -6,282 +6,796 @@
...
+loop:
+	for {
+		select {
+		case tgt, ok := <-outCh:
+			if !ok {
+				ss.log.Debugf("%s: out chan closed", ss)
+				// At this point we may still have an error to report,
+				// wait until error channel is emptied too.
+				outCh = nil
+				tr.safeCloseErrCh(ss)
```
marcoguerri, line 640: If `outCh` indicates that the test step is guaranteed to have returned, then whoever signaled completion by closing `outCh` should have closed also `errCh`. Responsibility of closing these channels should not be spread across multiple entities.

pkg/runner/test_runner.go
```diff
@@ -6,282 +6,796 @@
...
+			}
+		case <-ctx.Done():
+			ss.log.Debugf("%s: canceled 3, draining", ss)
+			for {
+				select {
+				case <-outCh:
+					break loop
+				case <-ss.errCh:
+					break loop
+				case <-time.After(tr.shutdownTimeout):
```
marcoguerri, line 662: This should not be necessary. It should be `stepRunner` guaranteeing to us that step termination is protected against `shutdownTimeout`. We shouldn't be doing it in multiple places.

pkg/runner/test_runner.go
```diff
@@ -6,282 +6,796 @@
...
+		ss := tr.steps[step]
+		tr.log.Debugf("monitor pass %d: current step %s", pass, ss)
+		// Check if all the targets have either made it past the injection phase or terminated.
+		ok := true
+		for _, ts := range tr.targets {
+			tr.log.Debugf("monitor pass %d: %s: %s", pass, ss, ts)
+			if ts.resCh == nil { // Not running anymore
+				continue
+			}
+			if ok && (ts.CurStep < step || ts.CurPhase < targetStepPhaseRun) {
```
marcoguerri, line 731: This truly deserves a comment. I don't understand this check, why we break down below, and why `if !ok`, we just wait for condition variable to be signaled and continue looping into the following step.

marcoguerri:
Tried to finish review of at least `test_runner.go`. Looks like I could not comment in #212 outside of the scope of that specific PR (which is really not ideal).

pkg/runner/test_runner.go
```diff
@@ -6,282 +6,796 @@
...
+		} else {
+			runErr = statectx.ErrPaused
+		}
+	default:
+		// We are not pausing and yet some targets were left in flight.
+		if len(inFlightTargets) > 0 {
+			ts := inFlightTargets[0]
+			runErr = &cerrors.ErrTestStepLostTargets{
+				StepName: tr.steps[ts.CurStep].sb.TestStepLabel,
+				Target:   ts.tgt.ID,
```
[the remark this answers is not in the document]
rojer9-fb: done

pkg/runner/test_runner.go
```diff
@@ -6,282 +6,796 @@
...
-// writeTargetWithResult attempts to deliver a Target on the input channel of a step,
-// returning the result of the operation on the result channel wrapped in the
-// injectionCh argument
-func (w *targetWriter) writeTargetWithResult(ctx context.Context, target *target.Target, ch injectionCh) {
-	err := w.writeTimeout(ctx, ch.stepIn, target, w.timeouts.StepInjectTimeout)
+func (tr *TestRunner) waitStepRunners(ctx statectx.Context) error {
+	tr.log.Debugf("waiting for step runners to finish")
+	swch := make(chan struct{})
+	go func() {
+		tr.mu.Lock()
```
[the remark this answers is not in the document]
rojer9-fb: this has to do with how waiting on `sync.Condition` works. when you want to wait for some internal state protected by a mutex to change (in this case `stepRunning` and `readerRunning`), you acquire a mutex, test the state and if it is not to your satisfaction, you invoke `tr.cond.Wait()` which releases the mutex and then suspends until signaled, at which point it re-locks and returns, letting you re-examine the state again. we signal `tr.cond` when things change (runner exits, rader exits of target handler exits).

pkg/runner/test_runner.go
```diff
@@ -6,282 +6,796 @@
...
+		defer tr.mu.Unlock()
+		// If there is a step with an error set, use that.
+		err = tr.checkStepRunners()
+		// If there isn't, enumerate ones that were still running at the time.
+		nrerr := &cerrors.ErrTestStepsNeverReturned{}
+		if err == nil {
+			err = nrerr
+		}
+		for _, ss := range tr.steps {
+			if ss.stepRunning {
```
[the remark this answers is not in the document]
rojer9-fb: the step runner context will eventually be canceled (via defer stepCancel())

pkg/runner/test_runner.go
```diff
@@ -6,282 +6,796 @@
...
+		// Injected successfully.
+		err := ss.ev.Emit(testevent.Data{EventName: target.EventTargetIn, Target: ts.tgt})
+		tr.mu.Lock()
+		ts.CurPhase = targetStepPhaseRun
+		if err != nil {
+			return fmt.Errorf("failed to report target injection: %w", err)
+		}
+		tr.mu.Unlock()
+		tr.cond.Signal()
+		if err != nil {
```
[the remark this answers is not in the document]
rojer9-fb: fixed

pkg/runner/test_runner.go
```diff
@@ -6,282 +6,796 @@
...
-func (w *targetWriter) writeTargetError(ctx context.Context, ch chan<- cerrors.TargetError, targetError cerrors.TargetError, timeout time.Duration) error {
+func (tr *TestRunner) injectTarget(ctx statectx.Context, ts *targetState, ss *stepState, log *logrus.Entry) error {
+	log.Debugf("%s: injecting into %s", ts, ss)
 	select {
+	case ss.inCh <- ts.tgt:
+		// Injected successfully.
+		err := ss.ev.Emit(testevent.Data{EventName: target.EventTargetIn, Target: ts.tgt})
+		tr.mu.Lock()
+		ts.CurPhase = targetStepPhaseRun
+		if err != nil {
```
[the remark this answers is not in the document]
rojer9-fb: you are right, fixed

pkg/runner/test_runner.go
rojer9-fb: the comment above explains the purpose of the entire loop - check if all the targets have either made it past the injection phase or terminated.
the first condition checks if the target has terminated or not (if terminated, there will not be a result channel).
this one checks if it made past the target injection phase.
it's perfectly fine for this to not be the case, in this case we just wait for things to change (for condition to be signaled).

i simplified the check a bit, since we break out on first !ok, there's no reason to test for it (a leftover from previous implementation where we'd iterate all targets anyway).

pkg/runner/test_runner.go
rojer9-fb: yes, except channels have different types and i couldn't find a way to cast them to something common.

pkg/runner/test_runner.go
rojer9-fb: this will be simplified when we have one channel for results (as discussed). i'll leave it as is for now.

PR #212, opened 2021-01-19
Remove unnecessary stepCancel

rojer:


marcoguerri:
This looks good, but I could not find a way to comment on code which is outside of the scope of this diff and I wanted to leave additional comments to the test runner, so I went back and commented in #208. Please have a look when you have a moment.

rojer9-fb:
will reopen with my facebook account

rojer9-fb:
#219

PR #213, opened 2021-01-19
Add ability to list jobs

rojer:
With filtering by state and tags.

ListJobs call is added to the Storage interface and implemented for both
in-memory and RDBMS storage plugins. RDBMS implementation requires
database schema changes to allow efficient lookups:
 * "state" column is added to the jobs table
 * "job_tags" table is added to map tags to job_id

Database migration script is provided to create and populate the new
columns. Going forward, they will be maintained by the server.

### DIFF
--- a/db/rdbms/migration/0003_add_jobs_state_column.sql
+++ b/db/rdbms/migration/0003_add_jobs_state_column.sql
@@ -1,0 +1,63 @@
+-- Copyright (c) Facebook, Inc. and its affiliates.
+--
+-- This source code is licensed under the MIT license found in the
+-- LICENSE file in the root directory of this source tree.
+
+-- +goose Up
+
+ALTER TABLE jobs ADD COLUMN state tinyint DEFAULT 0 AFTER extended_descriptor;
+CREATE INDEX job_state ON jobs (state, job_id);
+
+-- Populate the column.
+
+SET SQL_BIG_SELECTS=1;
+
+UPDATE
+  jobs
+INNER JOIN
+  ( -- This query retrieves job state event by its id.
+    SELECT
+      jobs.job_id AS job_id,
+      fe2.event_name AS state
+    FROM
+      jobs
+    INNER JOIN
+      ( -- This query finds id of the last job state event for each job.
+        SELECT
+          job_id,
+          MAX(event_id) AS max_event
+        FROM
+          framework_events fe
+        WHERE
+          fe.event_name IN ("JobStateStarted", "JobStateCompleted", "JobStateFailed",
+                            "JobStatePaused", "JobStatePauseFailed", "JobStateCancelling",
+                            "JobStateCancelled", "JobStateCancellationFailed")
+        GROUP BY
+          job_id
+      ) fe1
+    ON
+      fe1.job_id = jobs.job_id
+    INNER JOIN
+      framework_events fe2
+    ON
+      fe2.event_id = fe1.max_event
+  ) tt
+ON
+  jobs.job_id = tt.job_id
+SET
+  jobs.state =
+      -- Translate string to numeric representation.
+      IF(tt.state = "JobStateStarted", 1,
+      IF(tt.state = "JobStateCompleted", 2,
+      IF(tt.state = "JobStateFailed", 3,
+      IF(tt.state = "JobStatePaused", 4,
+      IF(tt.state = "JobStatePauseFailed", 5,
+      IF(tt.state = "JobStateCancelling", 6,
+      IF(tt.state = "JobStateCancelled", 7,
+      IF(tt.state = "JobStateCancellationFailed", 8,
+      0))))))));
+
+-- +goose Down
+
+DROP INDEX job_state ON jobs
+ALTER TABLE jobs DROP COLUMN state;
--- a/pkg/job/job.go
+++ b/pkg/job/job.go
@@ -6,6 +6,8 @@
 package job
 
 import (
+	"fmt"
+	"regexp"
 	"time"
 
 	"github.com/facebookincubator/contest/pkg/statectx"
@@ -62,6 +64,37 @@ type Job struct {
 	FinalReporterBundles []*ReporterBundle
 }
 
+type State int
+
+const (
+	JobStateUnknown            State = iota
+	JobStateStarted                  // 1
+	JobStateCompleted                // 2
+	JobStateFailed                   // 3
+	JobStatePaused                   // 4
+	JobStatePauseFailed              // 5
+	JobStateCancelling               // 6
+	JobStateCancelled                // 7
+	JobStateCancellationFailed       // 8
+)
+
+func (js State) String() string {
+	if js > 8 {
+		return fmt.Sprintf("JobState%d", js)
+	}
+	return []string{
+		"JobStateUnknown",
+		string(EventJobStarted),
+		string(EventJobCompleted),
+		string(EventJobFailed),
+		string(EventJobPaused),
+		string(EventJobPauseFailed),
+		string(EventJobCancelling),
+		string(EventJobCancelled),
+		string(EventJobCancellationFailed),
+	}[js]
+}
+
 // Cancel closes the cancel channel to signal cancellation
 func (j *Job) Cancel() {
 	j.StateCtxCancel()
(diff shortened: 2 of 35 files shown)


plugins/storage/rdbms/list_jobs.go
```diff
@@ -0,0 +1,81 @@
...
+	r.frameworkEventsLock.Unlock()
+	if err != nil {
+		return nil, fmt.Errorf("could not flush events before reading events: %v", err)
+	}
+
+	// Construct the query.
+	parts := []string{"SELECT jobs.job_id FROM jobs"}
+	// Tag filtering uses joins, 1 per tag.
+	for i := range query.Tags {
+		parts = append(parts, fmt.Sprintf(`INNER JOIN job_tags jt%d ON jobs.job_id = jt%d.job_id`, i, i))
```
tfg13, line 38: This is super super scary. We are one innocent regex change away from a full blown SQL injection in a read-only API call that does not even have ACLs of its own.

Would it maybe be possible to change this to append static strings `INNER JOIN job_tags ? ON jobs.job_id = ?` and keep a `[]interface{}` with the arguments in the same order to later pass it to `db.Query`? (Not entirely sure, not everything can be used with `?` (also discussed in the ticket you linked above).

tests/integ/common/storage.go
```diff
@@ -7,15 +7,21 @@ package common
...
-const DbURI = "contest:contest@tcp(mysql:3306)/contest_integ?parseTime=true"
+const DbURI = "contest:contest@tcp(localhost:3306)/contest_integ?parseTime=true"
+const DbURICI = "contest:contest@tcp(mysql:3306)/contest_integ?parseTime=true"
 
 func NewStorage(opts ...rdbms.Opt) (storage.Storage, error) {
-	return rdbms.New(DbURI, opts...)
+	if os.Getenv("CI") != "" {
+		return rdbms.New(DbURICI, opts...)
+	} else {
+		return rdbms.New(DbURI, opts...)
```
tfg13, line 23: Thanks for this, I replaced this string a hundred times already and every. single. time. I end up committing it by accident. Would you mind doing the same in `dblocker_test.go` as well?

tfg13 (commented):
Mostly nits, 2 main things:

- I am really scared by the SQL stuff, see inline
- Will the list API call return all jobs that ever ran and match the filters? I.e. will this return 100000s of jobs potentially? Should it? Do we know approx how many jobs until it OOMs or the DB aborts the query because it is too large?

db/rdbms/migration/0003_add_jobs_state_column.sql
```diff
@@ -0,0 +1,63 @@
...
+-- LICENSE file in the root directory of this source tree.
+
+-- +goose Up
+
+ALTER TABLE jobs ADD COLUMN state tinyint DEFAULT 0 AFTER extended_descriptor;
+CREATE INDEX job_state ON jobs (state, job_id);
+
+-- Populate the column.
+
+SET SQL_BIG_SELECTS=1;
```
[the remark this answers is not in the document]
rojer: it may take a while, for sure, but it won't explode the database as such.
fwiw, i tested with a dump of current xdb.contest and it takes about 20 seconds.

plugins/storage/rdbms/events.go
```diff
@@ -334,24 +335,37 @@ func (r *RDBMS) StoreFrameworkEvent(event frameworkevent.Event) error {
...
+
 // FlushFrameworkEvents forces a flush of the pending frameworks events to the database
 // Requires that the caller has already locked the corresponding buffer
 func (r *RDBMS) flushFrameworkEvents() error {
-
 	r.lockTx()
 	defer r.unlockTx()
 
-	insertStatement := "insert into framework_events (job_id, event_name, payload, emit_time) values (?, ?, ?, ?)"
+	// TODO: put this into a transaction.
```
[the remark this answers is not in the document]
rojer: for later - just pointing it out that it's possible now, in case anyone is reading this and wonders why the state field is not being used.

plugins/storage/rdbms/list_jobs.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/facebookincubator/contest/pkg/storage"
+	"github.com/facebookincubator/contest/pkg/types"
+)
+
+func (r *RDBMS) ListJobs(query *storage.JobQuery) ([]types.JobID, error) {
+	res := []types.JobID{}
+
+	// Quoting SQL strings is hard. https://github.com/golang/go/issues/18478
+	// For now, just disallow anything that is not [a-zA-Z0-9_-]
+	if err := job.CheckTags(query.Tags); err != nil {
```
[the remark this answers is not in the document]
rojer: done

plugins/storage/rdbms/list_jobs.go
rojer: oh, i totally agree and i went looking for a quoting function to do exactly what you are suggesting.
unfortunately, Go does not provide such interface in sql or sql driver interface, this is what https://github.com/golang/go/issues/18478 is about.
moreover, in general, it is not even possible to do it for Mysql with a simple function, as quoting converntion depends on connection settings! in particular, whether backslash escape sequences are interpreted or not. it's a pain.
so, in the end i settled for a restrictive regex as a means of ensuring there is no injection vector.

tests/integ/common/storage.go
rojer: done

rojer:
forgot to respond to this one, so here goes:

> Mostly nits, 2 main things:
> 
> * I am really scared by the SQL stuff, see inline

addressed

> * Will the list API call return all jobs that ever ran and match the filters? I.e. will this return 100000s of jobs potentially? Should it? Do we know approx how many jobs until it OOMs or the DB aborts the query because it is too large?

so, yes, the API has no pagination support at the moment. yes, unfiltered query will return all 100000 or so jobs, but for now it's just 100K integers, so it's not so bad. the first real use case for this will be to find all the paused jobs at the instance startup, of which there will never be more than a single instance can handle anyway, so it's fine.
i can add an explicit TODO to add pagination to API and put a hard limit on the number of returned results.

plugins/storage/rdbms/list_jobs.go
tfg13: Yeah I get the part about the quoting functions. I was just wondering if it was possible to construct the query with something like `INNER JOIN job_tags ? ON jobs.job_id = ?` and then pass the parameters into `db.Exec()`, as those are always properly escaped, automatically.

rojer: oh, yes, you are right. done

rojer:
rebased

plugins/storage/rdbms/list_jobs.go
tfg13: Did this change, I don't see it. Or maybe I misunderstood?

rojer: it did. there is still some Sprintf'ing happening there, but it's only for counters, all user-supplied data is passed via the `qargs` slice and substituted by the driver.

tfg13 (approved):

pkg/job/job.go
```diff
@@ -62,6 +64,44 @@ type Job struct {
...
+	case JobStateCancelling:
+		return string(EventJobCancelling)
+	case JobStateCancelled:
+		return string(EventJobCancelled)
+	case JobStateCancellationFailed:
+		return string(EventJobCancellationFailed)
+	}
+	return fmt.Sprintf("JobState%d", js)
+}
+
```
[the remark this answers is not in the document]
rojer: done

pkg/storage/job.go
```diff
@@ -52,6 +67,14 @@ func (jsm JobStorageManager) GetJobReport(jobID types.JobID) (*job.JobReport, er
 	return report, nil
 }
 
+func (jsm JobStorageManager) ListJobs(query *JobQuery) ([]types.JobID, error) {
+	res, err := storage.ListJobs(query)
+	if err != nil {
+		return nil, err
+	}
+	return res, nil
```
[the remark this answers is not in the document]
rojer: done

rojer:
addressed 2 comments i somehow missed before

PR #217, opened 2021-01-25
Fix run_tests.sh when running locally

fabriziocucci:
When running locally:
```
$ ./run_tests.sh
Re-running as root with sudo
./run_tests.sh: line 19: docker-compose: command not found﻿
```

Hopefully this doesn't break the CI run...


rojer9-fb (approved):
looks good, thanks!

PR #226, opened 2021-02-08
storage/rdbms: Ping database in init()

tfg13:
Otherwise the server won't notice if the DB URL doesn't lead anywhere and ends up with a nil storage engine in production


marcoguerri:
I was inclined to make storage interface implement `Ping`, and have `Ping` called in `SetStorage`. So that it's called for any storage, and the constructor `New` doesn't block. 

However, I also understand that it was discussed to change to remove `SetStorage` altogether, so there wouldn't be a central place where to run this check anymore, and  this would be the preferred approach.

marcoguerri (approved):

PR #227, opened 2021-02-11
use replica storage for status calls

mkelly121:
Added a read-only storage engine to be used in job status calls.

New methods `FetchAsync`, `GetJobReportAsync` and `GetJobRequestAsync` will use this read-only storage


cmds/contest/main.go
```diff
@@ -44,22 +44,43 @@ func main() {
...
 	if err != nil {
 		log.Warningf("Could not determine storage version: %v", err)
 	} else {
 		log.Infof("Storage version: %d", dbVer)
 	}
 
+	// replica storage initialization
+	log.Infof("Using database URI for replica storage: %s", *flagDBURI)
+
+	replicaDBURI := *flagDBURI
```
tfg13, line 69: Drop a comment here that this is just pointing to the main database for now, but could be used to connect to a replica

cmds/contest/main.go
```diff
@@ -44,22 +44,43 @@ func main() {
...
+	if err := storage.SetStorageAsync(s); err != nil {
+		log.Fatalf("Could not set replica storage: %v", err)
+	}
+
+	dbVer, err = s.Version()
+	if err != nil {
+		log.Warningf("Could not determine storage version: %v", err)
+	} else {
+		log.Infof("Storage version: %d", dbVer)
+	}
```
tfg13, line 83: lets fetch the version and compare it to the main one. This is by no means a check that we actually talk to the same DB, but if the version numbers are different something is very wrong and we should exit immediately

pkg/jobmanager/jobmanager.go
```diff
@@ -60,7 +60,7 @@ type JobManager struct {
 	jobsMu sync.Mutex
 	jobsWg sync.WaitGroup
 
-	jobStorageManager storage.JobStorage
+	jobStorageManager storage.JobStorageManager
```
tfg13, line 63: accidental change?

pkg/storage/storage.go
```diff
@@ -56,3 +57,29 @@ func SetStorage(storageEngine Storage) error {
...
+	if storage == nil {
+		return nil, fmt.Errorf("no storage engine assigned")
+	}
+	return storage, nil
+}
+
+// SetStorage sets the desired storage engine for read-only events. Switching to a new
+// storage engine implies garbage collecting the old one, with possible loss of
+// pending events if not flushed correctly
+func SetStorageAsync(storageEngine Storage) error {
```
tfg13, line 72: Maybe name this `SetAsyncStorage`, also copy-paste in first word of the doc comment :)

tfg13 (changes requested):
This looks good! Just some minor changes left

pkg/storage/job.go
```diff
@@ -48,19 +52,22 @@ func (jsm JobStorageManager) StoreJobReport(report *job.JobReport) error {
...
 
-// GetJobReport fetches a job report to the storage layer
+// GetJobReport fetches a job report from the storage layer
 func (jsm JobStorageManager) GetJobReport(jobID types.JobID) (*job.JobReport, error) {
-	return storage.GetJobReport(jobID)
+	report, err := storage.GetJobReport(jobID)
+	if err != nil {
+		return nil, err
+	}
+	return report, nil
```
tfg13: isn't this exactly the same as:
```suggestion
	return storage.GetJobReport(jobID)
```

pkg/storage/job.go
```diff
@@ -48,19 +52,22 @@ func (jsm JobStorageManager) StoreJobReport(report *job.JobReport) error {
...
-// GetJobReport fetches a job report from the storage layer
+// GetJobReportAsync fetches a job report from the read-only storage layer
 func (jsm JobStorageManager) GetJobReportAsync(jobID types.JobID) (*job.JobReport, error) {
-	return storageAsync.GetJobReport(jobID)
-}
-
-// ListJobs returns list of job IDs matching the query
-func (jsm JobStorageManager) ListJobs(query *JobQuery) ([]types.JobID, error) {
-	return storage.ListJobs(query)
+	report, err := storageAsync.GetJobReport(jobID)
```
tfg13, line 66: same

tfg13 (approved):
one super minor nit, only if you want to

rojer9-fb:
just ran into this while rebasing - no objections in principle but the "Async" suffix for storage methods seems confusing - usually "FooAsync" means "do Foo in a non-blocking manner", not "obtain a possibly stale version of Foo".

tfg13:
Feel free to rename it, this was the best I could come up with on short notice, not married to the name :)

PR #228, opened 2021-02-11
test_runner: Unit test to highlight pipeline bug

tfg13:
The pipeline fails if the last step takes longer than the shutdown timeout.
This test triggers that condition. Interestingly, it can fail in two different ways!
A regular unit test failure like we saw testing, and also a full blown panic for senging on a closed channel :(


tfg13:
This is just to demonstrate the problem, no need to review/merge for now

rojer9-fb:
incorporated into #232

PR #229, opened 2021-02-11
revert switch to new job runner pipeline

tfg13:
This is our fault, we haven't reimported code for a bit and there are a bunch of problems with the new pipeline that we still have to iron out. A bunch of people are on PTO, so this attempts to revert all the pipeline, resume and job listing changes, without throwing away the leak checker and the statectx fixes.


tfg13:
I tested this overnight and everything is green and the runner is working fine again

marcoguerri (approved):

PR #230, opened 2021-02-22
Add ability to list jobs

rojer9-fb:
With filtering by state and tags.

ListJobs call is added to the Storage interface and implemented for both
in-memory and RDBMS storage plugins. RDBMS implementation requires
database schema changes to allow efficient lookups:
 * "state" column is added to the jobs table
 * "job_tags" table is added to map tags to job_id

Database migration script is provided to create and populate the new
columns. Going forward, they will be maintained by the server.

This is take 2, first was #213 but was reverted due to unrelated issue.
There are no substantial changes here.

### DIFF
--- a/cmds/clients/contestcli/main.go
+++ b/cmds/clients/contestcli/main.go
@@ -24,6 +24,15 @@ import (
 //
 // Get the status of a job whose ID is 10
 //   ./contestcli-http status 10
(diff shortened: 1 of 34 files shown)


tfg13 (approved):
Ok, lets try again :D

PR #231, opened 2021-02-26
Run tests multiple times

rojer9-fb:
To catch flaky ones

Substantially, this is #214, take 2

### DIFF
--- a/cmds/plugins/plugins.go
+++ b/cmds/plugins/plugins.go
@@ -6,6 +6,8 @@
 package plugins
 
 import (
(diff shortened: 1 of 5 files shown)


tfg13 (approved):
We had this before. Will this make things (much) slower?

rojer9-fb:
> We had this before. Will this make things (much) slower?

slower, but not much. i saved a bunch of times with optimizations before, now i'm spending that budget. we are back to approx 8 minutes per test run, it's not too bad.

PR #233, opened 2021-02-26
Add instance tag

rojer9-fb:
An instance tag allows multiple instances to share a database without
stepping on each other's toes.

If `-instanceTag=XXX` is specified at server startup, the server will
limit its actions to jobs that have that tag, specifically:
 * Any job started through that instance will have this tag added (if not already present).
 * Listing jobs will only include jobs that have this tag.
 * Server will only return status of jobs with this tag.

### DIFF
--- a/cmds/contest/main.go
+++ b/cmds/contest/main.go
@@ -36,12 +36,14 @@ var (
 	flagServerID       = flag.String("serverID", "", "Set a static server ID, e.g. the host name or another unique identifier. If unset, will use the listener's default")
 	flagProcessTimeout = flag.Duration("processTimeout", api.DefaultEventTimeout, "API request processing timeout")
 	flagTargetLocker   = flag.String("targetLocker", inmemory.Name, "Target locker implementation to use")
+	flagInstanceTag    = flag.String("instanceTag", "", "A tag for this instance. Server will only operate on jobs with this tag and will add this tag to the jobs it creates.")
 )
 
 func main() {
 	flag.Parse()
 	log := logging.GetLogger("contest")
 	log.Level = logrus.DebugLevel
+	logging.Debug()
 
 	pluginRegistry := pluginregistry.NewPluginRegistry()
 
@@ -126,6 +128,9 @@ func main() {
 	}
 	if *flagServerID != "" {
 		opts = append(opts, jobmanager.APIOption(api.OptionServerID(*flagServerID)))
+	}
+	if *flagInstanceTag != "" {
+		opts = append(opts, jobmanager.OptionInstanceTag(*flagInstanceTag))
 	}
 
 	jm, err := jobmanager.New(&listener, pluginRegistry, opts...)
--- a/cmds/contest/main.go
+++ b/cmds/contest/main.go
@@ -22,6 +22,7 @@ import (
 	"github.com/facebookincubator/contest/pkg/target"
 
 	"github.com/facebookincubator/contest/plugins/listeners/httplistener"
+	"github.com/facebookincubator/contest/plugins/storage/memory"
 	"github.com/facebookincubator/contest/plugins/storage/rdbms"
 
 	"github.com/facebookincubator/contest/plugins/targetlocker/dblocker"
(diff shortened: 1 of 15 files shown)


tfg13:
Isn't this what the serverID was originally intended for? Or am I mixing things CC @insomniacslk

insomniacslk:
> Isn't this what the serverID was originally intended for? Or am I mixing things CC @insomniacslk

yes, part of the intent of ServerID was to associate a server identifier (host name, hash, whatever) to an operation, including what ends up in the DB. However the tagging, if not instance-specific, could be a useful addition, e.g. when dealing with multiple instances of the same service where the ConTest server is replicated. If so, I would do this in two steps, server ID (similar to this instance tag) and tags (e.g. "MyServiceGroup", "generic_instances", "supersecure" etc).

Note that currently *jobs* have a tags option in the job descriptor, but that's currently unused and not captured in the DB schema. We may want to pivot on the concept of tagging in multiple areas

tfg13:
> Note that currently _jobs_ have a tags option in the job descriptor, but that's currently unused and not captured in the DB schema. We may want to pivot on the concept of tagging in multiple areas

I think this changed and they are in the DB now and properly queryable (https://github.com/facebookincubator/contest/blob/master/db/rdbms/migration/0004_add_job_tags_table.sql)

rojer9-fb:
i think knowing the exact identity of the server that performed an operation is useful - for debugging, in particular.
but the purpose of instance label is different and i don't think server id should be used for that purpose.
@insomniacslk actually, with instance tag my intention is implementing exactly what you describe as the second step:
> tags (e.g. "MyServiceGroup", "generic_instances", "supersecure" etc).

my intention is to run default, purgatory and slc instances with `-instanceTag=default`, `purgatory` and `SLC` respectively. hwtest instance has its own db so it doesn't matter, but probably won't hurt either.

pkg/jobmanager/start.go
```diff
@@ -16,7 +17,28 @@ import (
...
+		found := false
+		for _, tag := range jd.Tags {
+			if tag == jm.config.instanceTag {
+				found = true
+				break
+			}
+		}
+		if !found {
+			jd.Tags = append(jd.Tags, jm.config.instanceTag)
+		}
```
tfg13, line 35: will this explode if a user submits a job with a tag that is also used as an instance tag of some other instance?

pkg/jobmanager/status.go
```diff
@@ -45,6 +37,32 @@ func (jm *JobManager) status(ev *api.Event) *api.EventResponse {
...
+	if jm.config.instanceTag != "" {
+		found := false
+		for _, tag := range currentJob.Tags {
+			if tag == jm.config.instanceTag {
+				found = true
+				break
+			}
+		}
+		if !found {
+			evResp.Err = fmt.Errorf("job %d belongs to a different instance, this is %q",
```
tfg13, line 50: This is a behaviour change, up until now any instance will happily send a job report. But I agree this is a good change, instances might not even run the exact same version or be unable to resolve some plugins

cmds/contest/main.go
```diff
@@ -35,56 +36,72 @@ var (
...
 	flagProcessTimeout = flag.Duration("processTimeout", api.DefaultEventTimeout, "API request processing timeout")
 	flagTargetLocker   = flag.String("targetLocker", inmemory.Name, "Target locker implementation to use")
+	flagInstanceTag    = flag.String("instanceTag", "", "Mandatory tag for this instance. Server will only operate on jobs with this tag and will add this tag to the jobs it creates.")
 )
 
 func main() {
 	flag.Parse()
 	log := logging.GetLogger("contest")
 	log.Level = logrus.DebugLevel
+	logging.Debug()
```
tfg13, line 46: on purpose or leftover from testing?

tfg13 (commented):
I am a bit concerned about mixing the instance tag with normal tags, which are user-controllable, right?

More generally speaking, one thing I learned during my big revert is that mixing multiple things in one commit can really hurt later. From what I can see this PR does a lot of things that are unrelated (not sure, are they?):

- Add support for the inmemory DB
- Make slowecho an internal test plugin that can no longer be used (not backwards compatible with old job configs)

pkg/jobmanager/start.go
rojer9-fb: it won't immediately explode and in fact, at present, nothing at all will happen. however, in the future, their paused job may be resumed by another instance which will probably not go well.

cmds/contest/main.go
rojer9-fb: on purpose. i understand that the intention of the line emmediately above is to log at debug level but we actually don't. this makes server log at debug level.

rojer9-fb:
> I am a bit concerned about mixing the instance tag with normal tags, which are user-controllable, right?

good point. i don't want to invent entire separate mechanism though, so maybe we can introduce concept of "internal" tags, e.g, those starting with `_` that cannot be added or removed via API?

> 
> More generally speaking, one thing I learned during my big revert is that mixing multiple things in one commit can really hurt later. From what I can see this PR does a lot of things that are unrelated (not sure, are they?):
> 
> * Add support for the inmemory DB

yes, if dbURI is empty we start with inmemory DB. purely for convenience and quick testing.

> * Make slowecho an internal test plugin that can no longer be used (not backwards compatible with old job configs)

right, it's an internal plugin used by job manager test so i put it under test plugins.

i can pull these changes out into separate commits.

tfg13:
> > I am a bit concerned about mixing the instance tag with normal tags, which are user-controllable, right?
> 
> good point. i don't want to invent entire separate mechanism though, so maybe we can introduce concept of "internal" tags, e.g, those starting with `_` that cannot be added or removed via API?

Yeah something like that is probably a good idea. Underscore sounds good to me.

> i can pull these changes out into separate commits.

If it is quick and easy to do, please do it, that makes the PRs so much nicer to read :)

rojer9-fb:
@tfg13 i added concept of internal tags and require that -instanceTag be an internal one (start with `_`). we allow internal tags in list filters, there's no particular reason not to. on the other hand, we do not allow them when creating jobs.

i also split the inmemory storage stuff into a separate commit.

cmds/contest/main.go
```diff
@@ -35,56 +36,72 @@ var (
 	flagServerID       = flag.String("serverID", "", "Set a static server ID, e.g. the host name or another unique identifier. If unset, will use the listener's default")
 	flagProcessTimeout = flag.Duration("processTimeout", api.DefaultEventTimeout, "API request processing timeout")
 	flagTargetLocker   = flag.String("targetLocker", inmemory.Name, "Target locker implementation to use")
+	flagInstanceTag    = flag.String("instanceTag", "", "Mandatory tag for this instance. Server will only operate on jobs with this tag and will add this tag to the jobs it creates.")
```
tfg13, line 39: Last nit: It says mandatory, what happens if this is not set?

tfg13 (approved):

cmds/contest/main.go
rojer9-fb: hm. actually, that's not what i meant. the flag is not mandatory, the tag on the jobs is mandatory if -instanceTag is specified. rephrased the description.

if the flag is not specified then instance will operate on all the jobs in the database with no additional filtering.

PR #240, opened 2021-03-13
Feature/contextual logger

xaionaro:
For easier review I splitted this change to 4 commits:
* https://github.com/facebookincubator/contest/pull/240/commits/6acb9d833938926af5f955ed8009259bef3f3550  --  a non-modified code, which we already reviewed and are using internally for another service, **a review seems redundant**
* https://github.com/facebookincubator/contest/pull/240/commits/a9c438aa2f9439fa1ae093b8ba54936d0a81d3f9  --  merging `statectx` and `xcontext`, **recommended to review**.
* https://github.com/facebookincubator/contest/pull/240/commits/a8942067166276488b065c9ef9568b2aa9f377f8  -- deleting `statectx`, and just renaming the import path `statectx` to `xcontext` everywhere, **a review seems redundant**.
* https://github.com/facebookincubator/contest/pull/240/commits/4fbb2cc8e1fa7bf4712babe4c2f92d0a724dc31e -- propagating new logging into everywhere, **recommended to review**.

---

# Justification

## Conceptual justification

The project is currently goes through the same problems which a lot of projects goes while getting more mature: logs became messy, flow management became messy, introducing of new features (like tracing) is difficult and so on. And here is a solution to solve these problems.

#### Logs

* Now we have a contextual logger. In particular it means:
    * we can add a field somewhere and it will be recursively reused in the child flows, and as consequence logs are much more structured (while natural using of the logger);
    * we have flexible log-level control (previously we had a single log-level for everything);
* ConTest was intended to be pluggable solution. And now the backend-logger is pluggable. For example in the current code it is possible to switch to `zap` from `logrus` by changing a couple of lines.
* Simplified log-fields control ([an example](https://github.com/facebookincubator/contest/pull/240/files#r594367915)).
* Now we have a TraceID and we can connect client and server logs (but we do not do it, yet; didn't want to make PR bigger).

#### Other

* Previously we had `statectx` to mitigate some flow issues. Here we basically do the same thing, but in a more generic and Go-idiomatic way. For example we preserve compatibility with standard package `context` (for example we preserved `WithCancel`, `WithTimeout` and so on). Thus, for example, there's no need to mix different types of contexts anymore.
* Now we have tracer and metrics (with structured tags) for basically free, which will help us to investigate timeout problems.

#### This is only an interface

Basically `Context` and `Logger` are just interfaces (which defines things, already required by ConTest) and adapters to implement these interfaces (in particular based on logrus and zap).

---

Logs before: https://paste.ubuntu.com/p/25JkqRb9PD/
Logs after (if the plain-text variant is picked): https://paste.ubuntu.com/p/DbkGKhMjBj/
Logs after (if `bundles.OptionLogFormat(bundles.LogFormatJSON)` is added): https://paste.ubuntu.com/p/H9HgH5qqw3/

As you can see, for example now `job_id` is propagated.


rojer9-fb:
can you share an example of log of a job run, before and after the change?

rojer9-fb:
it's a pretty big change, you should accompany it with a note on what it allows, what are practical aspects for code contributors, etc. if it exists internally, external version is still needed.

rojer9-fb:
it would be nice to have context-level methods for logging, so instead of `ctx.Logger().Infof()` one could simply `ctx.Infof()`, for brevity. i don't think it's unreasonable for context interface to have those methods.

pkg/runner/test_runner2.go
```diff
@@ -119,37 +116,37 @@ func (tr *TestRunner) Run(
...
+) (json.RawMessage, error) {
+
+	ctx = ctx.WithFields(xcontext.Fields{
+		"job_id": jobID,
+		"run_id": runID,
+	})
+
+	ctx.Logger().Debugf("== test runner starting job %d, run %d", jobID, runID)
+	resumeState, err := tr.run(ctx.WithTag("phase", "run"), t, targets, jobID, runID, resumeState)
+	ctx.Logger().Debugf("== test runner finished job %d, run %d, err: %v", jobID, runID, err)
```
xaionaro, line 129: An example how fields control is simplified.

pkg/runner/test_runner2.go
```diff
@@ -116,40 +113,40 @@ const resumeStateStructVersion = 2
...
+) (json.RawMessage, error) {
+
+	ctx = ctx.WithFields(xcontext.Fields{
+		"job_id": jobID,
+		"run_id": runID,
+	})
+
+	ctx.Debugf("== test runner starting job %d, run %d", jobID, runID)
+	resumeState, err := tr.run(ctx.WithTag("phase", "run"), t, targets, jobID, runID, resumeState)
+	ctx.Debugf("== test runner finished job %d, run %d, err: %v", jobID, runID, err)
```
xaionaro, line 129: An example of how fields control is simplified.

xaionaro:
> can you share an example of log of a job run, before and after the change?

Added links to the description.

> it's a pretty big change, you should accompany it with a note on what it allows, what are practical aspects for code contributors, etc. if it exists internally, external version is still needed.

Added some highlights to the description.

> it would be nice to have context-level methods for logging, so instead of `ctx.Logger().Infof()` one could simply `ctx.Infof()`, for brevity. i don't think it's unreasonable for context interface to have those methods.

Fixed.

tests/common/goroutine_leak_check/goroutine_leak_check.go
```diff
@@ -144,6 +144,10 @@ func LeakCheckingTestMain(m *testing.M, funcWhitelist ...string) {
 	ret := m.Run()
 	if ret == 0 {
 		time.Sleep(20 * time.Millisecond) // Give stragglers some time to exit.
+		runtime.Gosched()
+		runtime.GC()
+		runtime.Gosched()
+		runtime.GC()
```
tfg13, line 150: Why was this added?

tfg13 (approved):
Thanks for making this straightforward to review. I read all the contest changes, but only skimmed over the xcontest ones, as we already reviewed them elsewhere.

tests/common/goroutine_leak_check/goroutine_leak_check.go
xaionaro: Added a comment.

PR #255, opened 2021-03-30
dblocker: Avoid table locks, use batching

rojer9-fb:
* Do not lock DB rows upfront. When run against a nonexistent row it
   locks entire table, we don't want that.
 * Remove UPDATE statements. Without row locks that woudl be racy, so
   instead we use DELETE + INSERT. Note that this effectively deprecates
   the "valid" field and has the nice effect of keeping fewer stale rows
   around.
 * Use batching - insert and delete in batches of up to 100 targets,
   should reduce the number of roundtrips.


tfg13 (approved):
LGTM, let's try this approach!

PR #256, opened 2021-03-30
Add a dependency on github.com/u-root/u-root/tools/checklicenses

rojer9-fb:


tfg13:
Why does it say incompatible in go.mod now?

rojer9-fb:
`ENOIDEA`. that's exactly the entry `go get` adds, `go mod tidy` is now cool with it and doesn't get rid of it, run_lint works... dunno.

tfg13 (approved):

PR #257, opened 2021-03-31
bugfix(runner): Use jobctx inside jobrunner (to log job_id)

xaionaro:
One more fix after clashing big long job resumption PR with big long xcontext PR.

The problem was that `runCtx` was derived not from `jobCtx`, but just from `ctx`. Therefore `runCtx` had not contained `job_id`.

### DIFF
--- a/pkg/jobmanager/start.go
+++ b/pkg/jobmanager/start.go
@@ -75,9 +75,9 @@ func (jm *JobManager) startJob(ctx xcontext.Context, j *job.Job, resumeState *jo
 	jm.jobsMu.Lock()
 	defer jm.jobsMu.Unlock()
 	jobCtx, jobCancel := xcontext.WithCancel(ctx)
-	jobCtx2, jobPause := xcontext.WithNotify(jobCtx, xcontext.ErrPaused)
+	jobCtx, jobPause := xcontext.WithNotify(jobCtx, xcontext.ErrPaused)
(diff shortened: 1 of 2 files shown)


rojer9-fb (approved):

pkg/jobmanager/start.go
```diff
@@ -75,9 +75,9 @@ func (jm *JobManager) startJob(ctx xcontext.Context, j *job.Job, resumeState *jo
 	jm.jobsMu.Lock()
 	defer jm.jobsMu.Unlock()
 	jobCtx, jobCancel := xcontext.WithCancel(ctx)
-	jobCtx2, jobPause := xcontext.WithNotify(jobCtx, xcontext.ErrPaused)
+	jobCtx, jobPause := xcontext.WithNotify(jobCtx, xcontext.ErrPaused)
 	jm.jobs[j.ID] = &jobInfo{job: j, pause: jobPause, cancel: jobCancel}
-	go jm.runJob(jobCtx2, j, resumeState)
+	go jm.runJob(jobCtx.WithField("job_id", j.ID), j, resumeState)
```
rojer9-fb, line 80: this won't cover the resume case. why ont move it to runJob itself?

xaionaro: "resume job" also calls `startJob`, so it will cover. But you anyway right (seems more stable to further changes to move to `runJob`), so moved to `runJob`.

PR #263, opened 2021-04-08
Plugin API cleanup: remove Resume() and CanResume()

rojer9-fb:
These will not be used, don't need them.

The plan is to pass resume state directly to Run().

### DIFF
--- a/pkg/cerrors/cerrors.go
+++ b/pkg/cerrors/cerrors.go
@@ -10,17 +10,6 @@ import (
 	"strings"
 )
 
-// ErrResumeNotSupported indicates that a test step cannot resume. This can
-// be checked explicitly by the framework
-type ErrResumeNotSupported struct {
-	StepName string
-}
-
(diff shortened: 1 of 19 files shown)


xaionaro (approved):

xaionaro:
What is the expected behavior of `Run` if resumption is not implemented in a specific teststep plugin?

rojer9-fb:
successful pause is communicated by handling the pause signal and returning ErrPaused along with serialized state.
if the job was paused successfully, it will later get the same state and be expected to continue where it left off, i.e. produce results for targets that were in flight during pause and continue handling new targets.
if plugin doesn't support resume, it is not expected to do anything - it doesn;t need to handle pause, doesn't need to return anything special and will never be passed a non-nil resume state. jobs with such "naive" plugins will still support resume, but only between steps.

xaionaro:
OK, makes sense. Thanks :)

PR #264, opened 2021-04-15
Produce and commit run reports to storage as soon as the run ends

rojer9-fb:
Instead of waiting for the entire job to finish

Tested with unit tests and that "contestcli status" works with
intermediate state.

Status output after first run and after job finished:
https://gist.github.com/rojer9-fb/320074b037044bf242411f1134a06505

### DIFF
--- a/plugins/storage/memory/memory.go
+++ b/plugins/storage/memory/memory.go
@@ -9,6 +9,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"sort"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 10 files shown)


plugins/storage/memory/memory.go
```diff
@@ -184,32 +184,57 @@ func (m *Memory) GetJobRequest(_ xcontext.Context, jobID types.JobID) (*job.Requ
...
+		} else {
+			i := int(r.RunID) - 1
+			if i > len(jr.RunReports) {
+				ctx.Errorf("Incomplete set of run reports for job %d", jobID)
+				break
+			}
+			if i == len(jr.RunReports) {
+				jr.RunReports = append(jr.RunReports, nil)
+			}
+			jr.RunReports[i] = append(jr.RunReports[i], r)
```
tfg13, line 234: it took me a while to wrap my head around the double arrays

tfg13 (approved):

tfg13 (approved):

plugins/storage/memory/memory.go
rojer9-fb: yeah, me too

PR #266, opened 2021-04-17
Add end to end tests, with real server and client

rojer9-fb:
Client and server binary are refactored to be reusable within the same process,
tests/e2e/e2e_test starts server on a random port, uses client to submit jobs and check status.
This is as close to using standalone binaries as possible but still provides coverage.

Two E2E tests are implemented:
 * A simple test that runs a single job in one go and checks its status.
 * A pause/resume test that starts a job and then stops server multiple
   times to force pause and resume, then verifies that all teh steps
   have been executed once and in the correct order.

### DIFF
--- a/pkg/jobmanager/jobmanager.go
+++ b/pkg/jobmanager/jobmanager.go
@@ -156,6 +156,7 @@ func (jm *JobManager) Run(ctx xcontext.Context, resumeJobs bool) error {
 		lErr := jm.apiListener.Serve(apiCtx, a)
 		ctx.Infof("Listener shut down successfully.")
 		errCh <- lErr
+		close(errCh)
 	}()
 
 loop:
@@ -182,6 +183,7 @@ loop:
 	}
 	// Stop the API (if not already)
 	jm.StopAPI()
+	<-errCh
 	// Wait for jobs to complete or for cancellation signal.
 	doneCh := ctx.Done()
 	for !jm.checkIdle(ctx) {
--- a/plugins/storage/rdbms/init.go
+++ b/plugins/storage/rdbms/init.go
@@ -124,6 +124,16 @@ func (r *RDBMS) Version() (uint64, error) {
 	return migrationlib.DBVersion(r.db)
 }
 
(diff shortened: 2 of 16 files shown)


tfg13 (changes requested):
Is the 7MB compiled client binary included on purpose? That will bloat the repo

plugins/storage/rdbms/init.go
```diff
@@ -124,6 +124,16 @@ func (r *RDBMS) Version() (uint64, error) {
 	return migrationlib.DBVersion(r.db)
 }
 
+// Reset wipes entire database contents. Used in tests.
+func (r *RDBMS) Reset() error {
```
tfg13, line 128: Scary stuff. Should we at least yell a bit, like https://github.com/facebookincubator/contest/blob/master/plugins/targetlocker/dblocker/dblocker.go#L343 ?

Also DELETE vs. TRUNCATE. (not sure, just wanted to mention it)

pkg/jobmanager/jobmanager.go
```diff
@@ -182,6 +183,7 @@ loop:
 	}
 	// Stop the API (if not already)
 	jm.StopAPI()
+	<-errCh
```
tfg13, line 186: This effectively waits for API shutdown, is that right?

rojer9-fb: correct

plugins/storage/rdbms/init.go
rojer9-fb: there is no context to yell to though, the storage api doesn't use context (yet).

rojer9-fb:
no, binary got there by accident, updated .gitignore to exclude.

plugins/storage/rdbms/init.go
rojer9-fb: changed to TRUNCATE

rojer9-fb:
PTAL

tfg13 (approved):

PR #267, opened 2021-04-21
Process Status async + Fix write in read-only Status query codepath

tfg13:
- `Status` processing in the server mainloop no longer blocks the mainloop. This is important since Status performs a bunch of DB queries that can take a while (depending on DB latency). All other calls are still sync, which makes reasoning about things like job ID allocation easier (they are also fast anyway).
- There was a rare condition where an accidental write could happen in the read-only Status codepath. Status uses a set of `*Async` named query functions that can (but don't have to) be backed by a separate database connection to a read-only replica (to reduce load on the primary instance and spread DB queries). One of the query function in Status was not using the ro backend, leading to potential writes in this codepath.

Note: There were complaints before about naming these functions `*Async`, suggestions for better names are still accepted :)

Test Plan:

I put a sleep for 5s in status.go, then ran something like `go run [...]contestcli status 1 & go run [...]contestcli status 1 & go run [...]contestcli status 1 & go run [...]contestcli status 1`.
This means the client is running in parallel, and server processing may or may not.

Before the change:
`0.81s user 1.23s system 9% cpu 21.154 total`

After the change:
`0.84s user 1.38s system 32% cpu 6.771 total`


pkg/jobmanager/jobmanager.go
```diff
@@ -100,6 +100,15 @@ func New(l api.Listener, pr *pluginregistry.PluginRegistry, opts ...Option) (*Jo
 }
 
 func (jm *JobManager) handleEvent(ctx xcontext.Context, ev *api.Event) {
+	// handle status asynchronously, keep rest sync
```
rojer9-fb, line 103: i don't think we need to be making the distinction. code is already written with the assumption that multiple requests can be in flight at the same time, that's why we have jobsMu for example.

so, i'd say `go jm.handleEvent' in Run is all we need.

however, we also need a way to drain all requests gracefully when stopping the API.

tfg13: I was considering doing just `go jm.handleEvent` but didn't want to risk turning enabling concurrency everywhere at once.

Re: Draining. The context will abort in-flight requests, right? Do you think we need a way to wait for all of them to finish?

rojer9-fb: > I was considering doing just `go jm.handleEvent` but didn't want to risk turning enabling concurrency everywhere at once.

i think there isn't much risk, i certainly was already under assumption that things can go in parallel. `JobManager.jobs` is the only bit of shared state and we already have `jobsMu` for that.

> The context will abort in-flight requests, right? Do you think we need a way to wait for all of them to finish?

We should still wait for all requests to properly fold before returning. A simple waitgroup should do.

tfg13: Ok everything is async now, and I tested the waiting with a long sleep in status (note the time jump).
```
^C[2021-04-22T16:27:41.711+01:00 D httplistener.go:235] Received server shut down request
[2021-04-22T16:27:41.711+01:00 I server.go:215] Signal "interrupt", canceling
[2021-04-22T16:27:41.711+01:00 I jobmanager.go:157] Listener shut down successfully.
[2021-04-22T16:27:59.213+01:00 I server.go:224] Exiting, <nil>
```

Want me to add messages ("Waiting for handler shutdown")?

PR #268, opened 2021-04-21
cli: Fix output in case of server error + report timeout properly

tfg13:
If the server returned any error, the client would not print it, but instead say "invalid JSON response", because it tried to unmarshal the empty response data.

Also fixing a minor problem where it did not report the correct timeout in the error message.

Test Plan: (provoking a timeout error in the server)
Before:
```
> contest % go run ./cmds/clients/contestcli status 14
Requesting URL http://localhost:8080/status with requestor ID 'contestcli-http'
  with params:
    jobID: [14]
    requestor: [contestcli-http]

The server responded with status 400 Bad Request
cannot decode json response: unexpected end of JSON input
exit status 1
```

Now:
```
> contest % go run ./cmds/clients/contestcli status 14
Requesting URL http://localhost:8080/status with requestor ID 'contestcli-http'
  with params:
    jobID: [14]
    requestor: [contestcli-http]

The server responded with status 400 Bad Request
{
 "ServerID": "",
 "Data": {
  "Status": null
 },
 "Err": "Status failed: time out waiting for response after 3s"
}
```

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -144,7 +144,7 @@ func (a *API) SendReceiveEvent(ev *Event, timeout *time.Duration) (*EventRespons
 	case resp = <-ev.RespCh:
 		return resp, nil
 	case <-time.After(to):
-		return nil, fmt.Errorf("time out waiting for response after %v", timeout)
+		return nil, fmt.Errorf("time out waiting for response after %v", to)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/api/api.go
```diff
@@ -144,7 +144,7 @@ func (a *API) SendReceiveEvent(ev *Event, timeout *time.Duration) (*EventRespons
 	case resp = <-ev.RespCh:
 		return resp, nil
 	case <-time.After(to):
-		return nil, fmt.Errorf("time out waiting for response after %v", timeout)
+		return nil, fmt.Errorf("time out waiting for response after %v", to)
```
tfg13, line 147: `timeout` is wrong, it might be nil. `to` is the right value

PR #269, opened 2021-04-21
Add Storage.Close method

rojer9-fb:
To ensure all pending events are flushed when server shuts down.

### DIFF
--- a/cmds/contest/server/server.go
+++ b/cmds/contest/server/server.go
@@ -81,15 +81,24 @@ func ServerMain(cmd string, args []string, sigs <-chan os.Signal) error {
 
 	pluginRegistry := pluginregistry.NewPluginRegistry(ctx)
 
+	var storageInstances []storage.Storage
+	defer func() {
+		for i, s := range storageInstances {
+			if err := s.Close(); err != nil {
+				log.Errorf("Failed to close storage %d: %v", i, err)
+			}
+		}
+	}()
+
 	// primary storage initialization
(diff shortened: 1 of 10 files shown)


cmds/contest/server/server.go
```diff
@@ -221,6 +231,14 @@ func ServerMain(cmd string, args []string, sigs <-chan os.Signal) error {
 
 	err = jm.Run(ctx, *flagResumeJobs)
 
+	target.SetLocker(nil)
+	if err := storage.SetStorage(nil); err != nil {
+		log.Errorf("Failed to flush storage")
+	}
+	if err := storage.SetAsyncStorage(nil); err != nil {
+		log.Errorf("Failed to flush async storage")
+	}
```
tfg13, line 240: From how I read this, this will never error with `nil` as a storage engine

rojer9-fb: yes, and this is fine: if user want to disable storage, they should be allowed to. we do it when tearing down tests.

tfg13: Yeah I get the disable usecase, but not sure what we are printing here, as the log lines can never be reached?

tfg13 (approved):

cmds/contest/server/server.go
rojer9-fb: oh yeah, but lint will complain if i don't check the return value :)
but i guess i can simplify it by using a sentinel

PR #271, opened 2021-04-25
[xcontext] Add implementations of prometheus and trafficstars metrics.

xaionaro:
The effect is isolated to xcontext itself (contest does not use metrics, yet). Therefore I guess a simple stamp is good enough :)

Here I add two implementations of Metrics for xcontext.


rojer9-fb:
this could use some more tests

xaionaro:
@rojer9-fb : Fixed. And fixed a lot of bugs in the process.

xaionaro:
I'm merging these changes today, because they are isolated to xcontext and should not be able to affect ConTest itself. The only risky line is here: https://github.com/facebookincubator/contest/pull/271/files#diff-b66f25f3e58eb8419a42dcf6af5391173e4e4faea653ee767e0c165a58389e0dR81

In theory there should not be any problems, but there will be, let's just revert this line.

rojer9-fb:
SGTM

PR #277, opened 2021-05-04
Continue refresh locks for paused jobs until server exits

rojer9-fb:
If one job takes longer to pause than other(s), locks on successfully
paused job's targets may expire before the server gets the chance to resume them.

This change centralizes lock refresh in a background routine that
refreshes locks for all the currently running jobs. As jobs complete,
they are rmeoved from the map. Paused jobs, however, are not, and thus their targets continue to be refreshed.

### DIFF
--- a/pkg/runner/job_runner.go
+++ b/pkg/runner/job_runner.go
@@ -23,13 +23,18 @@ import (
 	"github.com/facebookincubator/contest/pkg/xcontext"
 )
 
+// jobInfo describes jobs currently being run.
+type jobInfo struct {
+	jobID     types.JobID
+	targets   []*target.Target
+	jobCtx    xcontext.Context
+	jobCancel func()
+}
+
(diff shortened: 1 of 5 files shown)


pkg/runner/job_runner.go
```diff
@@ -374,6 +354,65 @@ func (jr *JobRunner) Run(ctx xcontext.Context, j *job.Job, resumeState *job.Paus
 	return nil, nil
 }
 
+func (jr *JobRunner) lockRefresher() {
+	interval := jr.targetLockDuration / 10 * 9
```
tfg13, line 358: There was a comment here before why this is `/ 10 * 9` , can we keep that?

tfg13 (approved):

pkg/runner/job_runner.go
rojer9-fb: done

PR #280, opened 2021-05-18
Allow targets in running phase if the step has been paused

rojer9-fb:


tfg13 (approved):

pkg/runner/job_runner.go
```diff
@@ -212,6 +203,21 @@ func (jr *JobRunner) Run(ctx xcontext.Context, j *job.Job, resumeState *job.Paus
 				runErr = jr.emitTargetEvents(runCtx, testEventEmitter, targets, target.EventTargetAcquired)
 			}
 
+			// Check for pause during target acquisition.
+			select {
+			case <-ctx.Until(xcontext.ErrPaused):
```
xaionaro, line 208: I guess it won't be used, but still: it is also possible to write it this way:
```go
if ctx.IsSignaledWith(xcontext.ErrPaused) {
     ...
}
```
instead of
```go
select {
case <-ctx.Until(xcontext.ErrPaused):
    ...
default:
}
```

rojer9-fb: ah, that's good to know, thanks

PR #285, opened 2021-06-14
upstream resumption helpers and port sleep test step

tfg13:
Upstreaming some helpers that make is easier to write plugins that are resumable.

`ForEachTargetWithResume` is the resumable equivalent of `ForEachTarget`. This plugin is a one stop shop and handles everything, the implementor just needs to save per-step data.


rojer9-fb:
LGTM!

PR #289, opened 2021-06-23
[jobmanager] Add status for all runs instead of just the latest one

mimir-d:
- add Status.RunStatuses as a list of all runs for a given job
- keep Status.RunStatus populated from RunStatuses as backwards compat
- refactor JobManager.status() to retrieve all runs
- remove JobRunner.GetCurrentRun and test since they're no longer needed
  for the status command

testing:

- test backward compat
```
% go run . status 3 | jq '.Data.Status.RunStatus | [.JobID,.RunID]'     
Requesting URL http://localhost:8080/status with requestor ID 'contestcli-http'
  with params:
    requestor: [contestcli-http]
    jobID: [3]

The server responded with status 200 OK
[
  3,
  3
]
```

- test new RunStatuses field
```
% go run . status 3 | jq '.Data.Status.RunStatuses | map([.JobID,.RunID])'
Requesting URL http://localhost:8080/status with requestor ID 'contestcli-http'
  with params:
    jobID: [3]
    requestor: [contestcli-http]

The server responded with status 200 OK
[
  [
    3,
    1
  ],
  [
    3,
    2
  ],
  [
    3,
    3
  ]
]
```


tfg13:
Still reading the code, but this definitive needs a test plan, and we need to fix the client since this is a breaking API change. I'll ask some folks if they think that is ok, but I suspect the answer is yes.

Surprisingly, computing can be a bit expensive at times (lots of DB roundtrips), so I wonder if in the future we need to allow selecting only a subset of runs to be returned. And if we do that, should this maybe be a map? (Just some thoughts, not saying we need to do this right away)

pkg/jobmanager/status.go
```diff
@@ -127,26 +131,12 @@ func (jm *JobManager) status(ev *api.Event) *api.EventResponse {
 		JobReport:   report,
 	}
 
-	// Fetch the ID of the last run that was started
-	runID, err := jm.jobRunner.GetCurrentRunAsync(ev.Context, jobID)
+	jobStatus.RunStatuses, err = jm.jobRunner.BuildRunStatuses(ctx, currentJob)
```
tfg13, line 134: This is the "sync" codepath (this is a terrible way to name it and I still accept better suggestions).

Previously, this was using `GetCurrentRunAsync`, which talks to a separate DB instance that can point to a replica. There might be replication lag, but that is ok for the status call. We should keep this behaviour, this was done because computing the Status can require a bunch of DB round trips, a replicate might be closer and faster, and it also removes load from the primary instance.

tfg13 (approved):

PR #296, opened 2021-08-02
New reporter that can upload a file to a AWS S3 bucket.

llogen:
Here i suggest a new reporter that is able to upload a specific file to a S3 bucket. That could be expand to other S3 services. That feature can be very useful if a test contains building a binary and that binary should be automatically uploaded.

### DIFF
--- a/plugins/reporters/uploadtestfile/uploadtestfile.go
+++ b/plugins/reporters/uploadtestfile/uploadtestfile.go
@@ -1,0 +1,163 @@
+package uploadtestfile
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/fs"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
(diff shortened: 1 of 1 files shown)


tfg13:
hey there, thanks for the PR!

You'll have to figure out the CLA to make the bot happy :)

Regarding your code - I was wondering, should/can this be a test step instead of a result reporter?
Just asking because it does not seem to depend on job results at all, and I can imagine a lot of test flows want to upload files. Then again, test steps can run per target - is this intended to only run once for all the DUTs in a tests?

llogen:
Hi, your welcome :)

I contacted cla@fb.com to make the bot happy. 
I was also thinking about creating this as a test step, but the idea was, that you can choose either you upload a file for each test using the RunReporter or just upload a file for the hole test using the FinalReporter. From my point of view that would not be possible as a test step. Please correct me if I am wrong!

tfg13:
No you are right, that makes sense. The RunReporter runs after all the tests steps are done. If you use more than one run (`Runs: 2` or more), then the FinalReporter will run after all the repetitions are done.

llogen:
Okay thank you! 
I am after the CLA signing.

tfg13:
Hey, I am afraid the bot is still not happy and I can't move this forward without :(

PR #297, opened 2021-08-12
Added contest-generator

insomniacslk:
contest-generator is a tool used to generate a `contest` executable with
a custom set of plugins. See README.md for instructions.


insomniacslk:
This is still a work in progress. I have to automatically generate cmds/contest/main.go via github actions, and validate the build.

For now this PR has:
* the implementation of `contest-generator`
* a new `cmds/contest/main.go` obtained by running `./contest-generator`
* removed `cmds/contest/server/` which is now entirely embedded in the `main.go`. This was probably used for testing, so I'm looking for usage and to replace it with the code generator

insomniacslk:
I see why the `server` package. I'll try to fix the e2e tests

insomniacslk:
Note: the listener plugin is not customizable yet, because there is no listener interface. I will add it in a later PR

rojer9-fb:
i'd suggest a YMAL config file instead of a custom txt, for easier future extensibility.

insomniacslk:
> i'd suggest a YMAL config file instead of a custom txt, for easier future extensibility.

The format could be something like:
```
[targetmanagers]
- github.com/facebookincubator/contest/plugins/targetmanagers/targetlist
- ...

[testfetchers]
- ...
```

Is this what you have in mind @rojer9-fb ?

rojer9-fb:
> Is this what you have in mind @rojer9-fb ?

yeah, pretty much

insomniacslk:
Done. The code is also a lot simpler now, thanks for the suggestion @rojer9-fb  !

cmds/contest-generator/main.go
```diff
@@ -0,0 +1,134 @@
...
+		return nil, fmt.Errorf("failed to read file: %w", err)
+	}
+	var cfg Config
+	if err := yaml.Unmarshal(data, &cfg); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal YAML: %w", err)
+	}
+	if len(cfg.TargetManagers) < 1 {
+		return nil, fmt.Errorf("want at least 1 targetmanager, got %d", len(cfg.TargetManagers))
+	}
+	if len(cfg.TargetManagers) < 1 {
```
rojer9-fb, line 71: all 4 conditions in this block are the same while messages are different. looks like copy-paste error.

cmds/contest-generator/main.go
```diff
@@ -0,0 +1,134 @@
...
+
+	data, err := ioutil.ReadFile(*flagTemplate)
+	if err != nil {
+		log.Fatalf("Failed to read template file '%s': %v", *flagTemplate, err)
+	}
+	t, err := template.New("contest").Funcs(funcMap).Parse(string(data))
+	if err != nil {
+		log.Fatalf("Template parsing failed: %v", err)
+	}
+	// additional plugin names from a text file, one line per plugin import
```
rojer9-fb, line 95: this comment is outdated

cmds/contest-generator/main.go
```diff
@@ -0,0 +1,134 @@
...
+	// path
+	fd, err := os.Open(*flagFromFile)
```
rojer9-fb, line 97: any reason not to use ioutil.ReadFile here?

cmds/contest-generator/README.md
```diff
@@ -0,0 +1,63 @@
...
+`contest-generator` is a tool used to build ConTest with the plugins you want.
+
+We provide a standard [main.go](/cmds/contest/main.go), where we only include
+the most common plugins.
+
+If you need to build `ConTest` with a custom set of plugins without maintaining
+your own `main.go` program, you can `contest-generator`.
+
+Just define your plugins in
+[core-plugins.txt](/cmds/contest-generator/core-plugins.txt) with the format
```
rojer9-fb, line 12: core-plugins.txt -> .yml, here and in a few other places

rojer9-fb (changes requested):

cmds/contest-generator/main.go
insomniacslk: It's my tendency to use `io.Reader` instead of `string` to make testing easier. However I didn't add testing to this module, and it's probably a good idea to do so at least for `parseConfig`

cmds/contest-generator/README.md
insomniacslk: whoops, thanks

cmds/contest-generator/main.go
insomniacslk: I also realized that I don't handle duplicate module names here, and I should. I'll add a check, and an optional `name` field where the user can specify an alternative import name that will result in
```
import_name "github.com/blah/duplicate_name
```

cmds/contest-generator/README.md
```diff
@@ -0,0 +1,63 @@
+## ConTest Generator
+
+`contest-generator` is a tool used to build ConTest with the plugins you want.
+
+We provide a standard [main.go](/cmds/contest/main.go), where we only include
+the most common plugins.
+
+If you need to build `ConTest` with a custom set of plugins without maintaining
+your own `main.go` program, you can `contest-generator`.
```
tfg13, line 9: I think there should be a verb somewhere here
```suggestion
your own `main.go` program, you use can `contest-generator`.
```

cmds/contest/main.go
```diff
@@ -1,8 +1,6 @@
-// Copyright (c) Facebook, Inc. and its affiliates.
-//
-// This source code is licensed under the MIT license found in the
-// LICENSE file in the root directory of this source tree.
```
tfg13, line 4: Won't the bot get angry about the missing copyright header?

cmds/contest/server/server.go
```diff
@@ -173,7 +244,7 @@ func ServerMain(cmd string, args []string, sigs <-chan os.Signal) error {
 	plugins.Init(pluginRegistry, ctx.Logger())
 
 	// spawn JobManager
-	listener := httplistener.NewHTTPListener(*flagListenAddr)
+	listener := httplistener.New(*flagListenAddr)
```
tfg13, line 247: Should (maybe in a future version) the listener plugin be selectable in the generator as well?

cmds/contest/server/server.go
```diff
@@ -61,7 +62,74 @@ func initFlags(cmd string) {
...
+// PluginConfig contains the configuration for all the plugins desired in a
+// server instance.
+type PluginConfig struct {
+	TargetManagers []TargetManagerConfig
+	TestFetchers   []TestFetcherConfig
+	TestSteps      []TestStepConfig
+	Reporters      []ReporterConfig
+}
+
+func registerPlugins(pluginRegistry *pluginregistry.PluginRegistry, pluginConfig *PluginConfig) error {
```
tfg13, line 99: Almost everything in this file already exists in `cmds/plugins/plugins.go` (which is the wrong place now that I think of it)

cmds/contest/server/server.go
```diff
@@ -173,7 +244,7 @@ func ServerMain(cmd string, args []string, sigs <-chan os.Signal) error {
 	plugins.Init(pluginRegistry, ctx.Logger())
```
tfg13, line 244: I think this initialises a hardcoded list of plugins again :/

tfg13 (changes requested):
See comments, I think this duplicates a lot of code and initialises some plugins twice.

The generator code lgtm.

Oh, and the commit contains a 15mb binary blob (server), please remove :)

cmds/contest/main.go
insomniacslk: it will, good catch

cmds/contest/server/server.go
insomniacslk: yes, see my comment above. I am leaving it out for now as we don't have an interface for listeners yet, but I plan on adding it after this PR, and to adjust the generator to include that. But I wouldn't pack all of this in a single PR

cmds/contest/server/server.go
insomniacslk: good point, will look into it

cmds/contest/server/server.go
insomniacslk: I'll see what I can reuse, can't remember now why I recreated it

insomniacslk:
Addressed feedback from Deomid, and added validation / duplicate checks. Will address Tobias's feedback in the next iteration

cmds/contest-generator/contest.go.template
```diff
@@ -0,0 +1,78 @@
+{{/* This file is the template source. The following comment obviously doesn't apply here */ -}}
+// This is a generated file, edits should be made in the corresponding source
+// file, and this file regenerated using
+//   `contest-generator --from core-plugins.txt`.
```
rojer9-fb, line 4: this is a template that will be used with other configs, so the command will be wrong for uses other than core-plugins.
i think the template should have a variable, like {{ .CommandLine }}

cmds/contest/main.go
insomniacslk: done

cmds/contest-generator/README.md
insomniacslk: done

cmds/contest-generator/README.md
insomniacslk: done

cmds/contest-generator/main.go
insomniacslk: done

cmds/contest-generator/main.go
insomniacslk: done

cmds/contest-generator/main.go
insomniacslk: done

cmds/contest-generator/main.go
insomniacslk: done

cmds/contest/server/server.go
insomniacslk: done

insomniacslk:
I need to address more feedback, and add testing

cmds/contest-generator/contest.go.template
insomniacslk: I thought of that comment just as an example, not as the actual command. Maybe I can replace it with "--from your-plugins-configuration.yml" to be more clear?

cmds/contest/server/server.go
insomniacslk: I'm going to resolve this

cmds/contest-generator/contest.go.template
insomniacslk: changed, let me know if that's OK

cmds/contest/server/server.go
insomniacslk: Why do you think it's a hardcoded list? `pluginRegistry` is initialized with `pluginregistry.NewPluginRegistry` (which has an empty list of plugins), and only later populated in `server.Main` from the plugins written in the generated code

cmds/contest-generator/contest.go.template
insomniacslk: ah I see. The generic name is also incorrect for `cmds/contest/main.go`

rojer9-fb: well, now it won't match anyone's config :)
i was thinking of something along the lines of:

```
// Generated with:
//  { .CommandLine }}
```
with
```
  CommandLine: strings.Join(or.Args, " ")
```

or some such

insomniacslk: done, with filename passed to the templaet

rojer9-fb:
LGTM, good to go once Tobias is also happy

insomniacslk:
split config functions and types into `config.go`

insomniacslk:
will add tests for configuration objects after lunch

insomniacslk:
Tests added. @tfg13 PTAL

cmds/contest/server/server.go
```diff
@@ -1,12 +1,8 @@
-// Copyright (c) Facebook, Inc. and its affiliates.
-//
-// This source code is licensed under the MIT license found in the
```
tfg13, line 3: copyright header deleted here as well

cmds/contest/server/server.go
tfg13: Maybe I am reading it wrong, but won't `plugins.Init` end up here: https://github.com/facebookincubator/contest/blob/884d78e8b2a2f852997226df0ad0e14573f0d549/cmds/plugins/plugins.go#L66

And then load that hardcoded list of plugins?
Btw that thing also handles user functions for templates, they are basically plugins there.

insomniacslk: ooh you're right. If I'm reading that correctly, we don't need to call `Init` at all, because we call `registerPlugins` with the plugins initialized in `main.go`. So just removing it should be enough?

cmds/contest-generator/contest.go.template
insomniacslk: @rojer do you think we need the full command line, or a basic `contest-generator --from $filename` is enough? For example, if I run it with `go run` to regenerate cmds/contest/main.go,  it will look like
```
/tmp/somewhere/contest-generator -o ../contest/main.go
```
which is less useful than
```
contest-generator --from core-plugins.yml
```
(the plugins config file is not part of the command line, it's taken from the defaults)

cmds/contest/server/server.go
insomniacslk: done

cmds/contest/server/server.go
insomniacslk: I have removed it, and it seems it's only loading the plugins I expect. It also seems that the `cmds/plugins` package is not used anywhere else? Maybe it could be removed altogether?

insomniacslk: except in `db/rdbms/migration/0002_migrate_descriptor_to_extended_descriptor.go` but if I understand correctly it's OK to remove that file, and who needs migrations will check out earlier commits?

insomniacslk:
fixed tests

mimir-d:
would you say that adding a build.sh script to do all of what the readme expects the user do to (go build stuff) is helpful?
I'm thinking since this PR has the semantics of providing a good to use binary with contest, maybe that goes the extra step for a turn-key solution.

insomniacslk:
> would you say that adding a build.sh script to do all of what the readme expects the user do to (go build stuff) is helpful?
> I'm thinking since this PR has the semantics of providing a good to use binary with contest, maybe that goes the extra step for a turn-key solution.

Not entirely sure if I got your point, but if you are referring to simplifying the process of building ConTest with contest-generator, there is already a pre-generated `main.go` under `cmds/contest`. I think a build script would only add more layers and confusion, and if the user needs to customize the build, they still need to edit a text file and call `contest-generator`, which is standard process for Go programs.

At the same time, this reminded me that `contest-generator` needs to be called by GitHub Actions, and rebuild and validate `cmds/contest/main.go`, which I'm going do to now

cmds/contest/server/server.go
```diff
@@ -170,10 +223,8 @@ func ServerMain(cmd string, args []string, sigs <-chan os.Signal) error {
 		log.Fatalf("Invalid target locker name %q", *flagTargetLocker)
 	}
 
-	plugins.Init(pluginRegistry, ctx.Logger())
```
tfg13, line 173: ok, this fixes loading plugins multiple times, but now we don't load the user functions at all anymore

mimir-d:
Yes, I'm talking about simplifying and guiding the user. The thing I have in mind is to have some extra checks for the user (I'm pretty sure a lot of people don't read the manual), like build.sh would point you to $EDITOR for the yaml file, then calls the contest-generator and go build and says "your binary is here".

cmds/contest/server/server.go
insomniacslk: good point, I'll re-add user function loading

insomniacslk:
> Yes, I'm talking about simplifying and guiding the user. The thing I have in mind is to have some extra checks for the user (I'm pretty sure a lot of people don't read the manual), like build.sh would point you to $EDITOR for the yaml file, then calls the contest-generator and go build and says "your binary is here".

Honestly I think this is too far. For that kind of users we should just provide prebuilt binaries

insomniacslk:
since `cmds/plugins` was removed because redundant, but still used by the migration tools, I moved that code directly into the migration tool

insomniacslk:
OK this is finally ready for the final review. A recap of the changes:

* added contest-generator (cmds/contest-generator/*)
* generated cmds/contest/main.go using contest-generator, which includes a comment with the commands to regenerate it
* tested contest-generator in GitHub actions so that it generates and validates the new code, compares it to cmds/contest/main.go, and fails if they are different (i.e. if the user forgot to regenerate it upon changes)
* removed the now redundant cmds/plugins/ and merged the remaining bits into the only consumer (the migration tool)
* included user documentation for contest-generator (cmds/contest-generator/README.md)
* provided default configuration file, with example for import aliasing (cmds/contest-generator/core-plugins.yml)
* tested parseConfig (cmds/contest-generator/config_test.go)
* renamed httplistener.NewHTTPListener to httplistener.New, waiting to create a listener interface in the next PRs (so that listeners will be configurable too - added as TODO)
* added TODO to include also user functions into code generation
* renamed cmds/contest/server.ServerMain to cmds/contest/server.Main (redundant prefix)
*

cmds/plugins/plugins.go
```diff
@@ -1,107 +0,0 @@
...
-
-	// Register Reporter plugins
-	for _, rfloader := range reporters {
-		if err := pluginRegistry.RegisterReporter(rfloader()); err != nil {
-			log.Fatalf("%v", err)
-		}
-	}
-
-	// user-defined function registration
-	testInitOnce.Do(func() {
```
tfg13, line 98: I think the test fails because you removed this `Once`, so on the second server start it tries to register the user functions again (they are not in the plugin registry, but in a global variable (boo!), so they should fail with "duplicate user function"). See below why this is not visible in the output

tests/e2e/e2e_test.go
```diff
@@ -104,7 +108,16 @@ func (ts *E2ETestSuite) startServer(extraArgs ...string) {
...
+		pc := server.PluginConfig{
+			TargetManagerLoaders: []target.TargetManagerLoader{
+				targetlist.Load,
+				targetlist_with_state.Load,
+			},
+			TestFetcherLoaders: []test.TestFetcherLoader{literal.Load},
+			TestStepLoaders:    []test.TestStepLoader{cmd.Load, sleep.Load},
+			ReporterLoaders:    []job.ReporterLoader{targetsuccess.Load, noop.Load},
+		}
+		_ = server.Main(&pc, "contest", args, serverSigs)
```
tfg13, line 120: You should not throw away the error here, this should have said "error duplicate user function" :)

cmds/plugins/plugins.go
insomniacslk: what to say, thanks for debugging it for me! Fix is coming right away

tests/e2e/e2e_test.go
insomniacslk: right. And it was already ignored in the original code, I just added `_ = ` to make the linter happy

insomniacslk: done, now the e2e tests are surfacing errors in the test report instead of ignoring them

cmds/plugins/plugins.go
insomniacslk: fixed

insomniacslk:
tests are green, and also improved error surfacing

tfg13 (approved):

PR #298, opened 2021-08-16
plugins/teststeps/uploadFile/uploadFile:

llogen:
Added a new teststep that can upload a specific file to an AWS S3 Bucket. This works compressed or not. The link will be printed in the report by the Stdout Event.

Signed-off-by: llogen <christophlange95@googlemail.com>

### DIFF
--- a/plugins/teststeps/uploadFile/uploadFile.go
+++ b/plugins/teststeps/uploadFile/uploadFile.go
@@ -1,0 +1,281 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+//
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+package uploadfile
+
+import (
+	"archive/tar"
+	"bytes"
+	"compress/gzip"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"strconv"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/facebookincubator/contest/pkg/event"
+	"github.com/facebookincubator/contest/pkg/event/testevent"
+	"github.com/facebookincubator/contest/pkg/target"
+	"github.com/facebookincubator/contest/pkg/test"
+	"github.com/facebookincubator/contest/pkg/xcontext"
+	"github.com/facebookincubator/contest/plugins/teststeps"
+	"github.com/prometheus/common/log"
+)
+
+// Define some constants for uploading things into a S3 bucket
+const (
+	S3_REGION = "" // your aws server region
+	S3_BUCKET = "" // your aws bucket
+	S3_PATH   = "" // your path where to upload the file
+)
+
+// Name is the name used to look this plugin up.
+var Name = "UploadFile"
+
+// Event names for this plugin.
+const (
+	EventCmdStdout = event.Name("CmdStdout")
+)
+
+// Events defines the events that a TestStep is allow to emit
+var Events = []event.Name{
+	EventCmdStdout,
+}
+
+// Cmd is used to run arbitrary commands as test steps.
+type UploadFile struct {
+	path       *test.Param // Path to file that shall be uploaded
+	filename   *test.Param // Filename to file that shall be uploaded
+	emitStdout bool        // Bool that defines that Stdout Event shall be triggered
+	compress   bool        // Bool that defines if the data should be compressed before uploading it
+}
+
+// Datatype for the emmiting the Stdout Event
+type eventCmdStdoutPayload struct {
+	Msg string
+}
+
+// Name returns the plugin name.
+func (ts UploadFile) Name() string {
+	return Name
+}
+
+func emitEvent(ctx xcontext.Context, name event.Name, payload interface{}, tgt *target.Target, ev testevent.Emitter) error {
+	payloadStr, err := json.Marshal(payload)
+	if err != nil {
+		return fmt.Errorf("cannot encode payload for event '%s': %v", name, err)
+	}
+	rm := json.RawMessage(payloadStr)
+	evData := testevent.Data{
+		EventName: name,
+		Target:    tgt,
+		Payload:   &rm,
+	}
+	if err := ev.Emit(ctx, evData); err != nil {
+		return fmt.Errorf("cannot emit event EventCmdStart: %v", err)
+	}
+	return nil
+}
+
+// Run executes the cmd step.
+func (ts *UploadFile) Run(ctx xcontext.Context, ch test.TestStepChannels, params test.TestStepParameters,
+	ev testevent.Emitter, resumeState json.RawMessage) (json.RawMessage, error) {
+
+	if err := ts.validateAndPopulate(params); err != nil {
+		return nil, err
+	}
+	f := func(ctx xcontext.Context, target *target.Target) error {
+		// expand args
+		path, err := ts.path.Expand(target)
+		if err != nil {
+			return fmt.Errorf("failed to expand argument '%s': %v", ts.path, err)
+		}
+		filename, err := ts.filename.Expand(target)
+		if err != nil {
+			return fmt.Errorf("failed to expand argument dir '%s': %v", ts.filename, err)
+		}
+		// Compress if compress parameter is true
+		if ts.compress {
+			// Create output file
+			newpath := path + ".tar.gz"
+			tar_gz, err := os.Create(newpath)
+			if err != nil {
+				return fmt.Errorf("error writing archive: %w", err)
+			}
+			defer tar_gz.Close()
+			// Create the archive and write the output to the "out" Writer
+			err = createTarArchive(path, tar_gz)
+			if err != nil {
+				log.Fatalln("Error creating archive:", err)
+			}
+			fmt.Println("Tar archive created successfully")
+			path = newpath
+			filename = filename + ".tar.gz"
+		}
+		// Upload file
+		url, err := upload(path, filename)
+		fmt.Println(url)
+		if err != nil {
+			return err
+		}
+		if ts.emitStdout {
+			log.Infof("Emitting stdout event")
+			if err := emitEvent(ctx, EventCmdStdout, eventCmdStdoutPayload{Msg: url}, target, ev); err != nil {
+				log.Warnf("Failed to emit event: %v", err)
+			}
+		}
+		return nil
+	}
+	return teststeps.ForEachTarget(Name, ctx, ch, f)
+}
+
+// Retrieve all the parameters defines through the jobDesc
+func (ts *UploadFile) validateAndPopulate(params test.TestStepParameters) error {
+	// validate path and filename
+	ts.path = params.GetOne("path")
+	if ts.path.IsEmpty() {
+		return errors.New("invalid or missing 'path' parameter, must be exactly one string")
+	}
+	ts.filename = params.GetOne("filename")
+	if ts.filename.IsEmpty() {
+		return errors.New("invalid or missing 'filename' parameter, must be exactly one string")
+	}
+	// validate emit_stdout
+	emitStdoutParam := params.GetOne("emit_stdout")
+	if !emitStdoutParam.IsEmpty() {
+		v, err := strconv.ParseBool(emitStdoutParam.String())
+		if err != nil {
+			return fmt.Errorf("invalid non-boolean `emit_stdout` parameter: %v", err)
+		}
+		ts.emitStdout = v
+	}
+	// validate compress
+	compressParam := params.GetOne("compress")
+	if !compressParam.IsEmpty() {
+		v, err := strconv.ParseBool(compressParam.String())
+		if err != nil {
+			return fmt.Errorf("invalid non-boolean `emit_stdout` parameter: %v", err)
+		}
+		ts.compress = v
+	}
+	return nil
+}
+
+// ValidateParameters validates the parameters associated to the TestStep
+func (ts *UploadFile) ValidateParameters(_ xcontext.Context, params test.TestStepParameters) error {
+	if params.Get("path") == nil {
+		return fmt.Errorf("could not validate parameter")
+	}
+	return nil
+}
+
+// New initializes and returns a new Cmd test step.
+func New() test.TestStep {
+	return &UploadFile{}
+}
+
+// Load returns the name, factory and events which are needed to register the step.
+func Load() (string, test.TestStepFactory, []event.Name) {
+	return Name, New, Events
+}
+
+func createTarArchive(file string, buf io.Writer) error {
+	// Create gzip and tar writers
+	gzwriter := gzip.NewWriter(buf)
+	defer gzwriter.Close()
+	tarwriter := tar.NewWriter(gzwriter)
+	defer tarwriter.Close()
+	// Write file into tar archive
+	err := addFileToArchive(tarwriter, file)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func addFileToArchive(tarwriter *tar.Writer, filename string) error {
+	// Open the file which shall be written into the tar archive
+	file, err := os.Open(filename)
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+	// Retrieve the file stats
+	info, err := file.Stat()
+	if err != nil {
+		return err
+	}
+	// Create a tar header from the file stats
+	header, err := tar.FileInfoHeader(info, info.Name())
+	if err != nil {
+		return err
+	}
+	header.Name = filename
+	// Write file header to the tar archive
+	err = tarwriter.WriteHeader(header)
+	if err != nil {
+		return err
+	}
+	// Write the file into the tar archive
+	_, err = io.Copy(tarwriter, file)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// Upload the file that is specified in the JobDescritor
+func upload(path string, filename string) (string, error) {
+	// Read the file that should be uploaded
+	bodyBytes, err := ioutil.ReadFile(path)
+	if err != nil {
+		return "Could not read the file.", err
+	}
+
+	// Create a single AWS session (we can re use this if we're uploading many files)
+	s, err := session.NewSession(&aws.Config{Region: aws.String(S3_REGION),
+		Credentials: credentials.NewSharedCredentials(
+			"", // your credential file path (default if empty)
+			"", // profile name (default if empty)
+		)})
+	if err != nil {
+		return "Could not open a new session.", err
+	}
+
+	currentTime := time.Now()
+
+	// Config settings: this is where you choose the bucket, filename, content-type etc.
+	// of the file you're uploading.
+	fileName := fmt.Sprintf("%s/%s_%s", S3_PATH, currentTime.Format("20060102_150405"), filename)
+
+	// Uploading the file
+	_, err = s3.New(s).PutObject(&s3.PutObjectInput{
+		Bucket:               aws.String(S3_BUCKET),
+		Key:                  aws.String(fileName),
+		ACL:                  aws.String("public-read"),
+		Body:                 bytes.NewReader(bodyBytes),
+		ContentLength:        aws.Int64(int64(len(bodyBytes))),
+		ContentType:          aws.String(http.DetectContentType(bodyBytes)),
+		ContentDisposition:   aws.String("attachment"),
+		ServerSideEncryption: aws.String("AES256"),
+	})
+	if err != nil {
+		return "Could no upload the file.", err
+	} else {
+		fmt.Printf("Pushed the file to S3 Bucket! \n")
+	}
+	// Create download link for public ACL
+	url := "https://" + S3_BUCKET + ".s3.amazonaws.com/" + fileName
+
+	return url, err
+}


walterchris:
I would rename the plugin to something that fits better, like awsFileUpload or so, as this does not upload files in general - but rather to AWS. Or integrate AWS as an option.

plugins/teststeps/uploadFile/uploadFile.go
```diff
@@ -0,0 +1,281 @@
...
+		}
+		// Upload file
+		url, err := upload(path, filename)
+		fmt.Println(url)
+		if err != nil {
+			return err
+		}
+		if ts.emitStdout {
+			log.Infof("Emitting stdout event")
+			if err := emitEvent(ctx, EventCmdStdout, eventCmdStdoutPayload{Msg: url}, target, ev); err != nil {
```
walterchris, line 133: Can't you use a custom event here?

plugins/teststeps/uploadFile/uploadFile.go
```diff
@@ -0,0 +1,281 @@
...
+			filename = filename + ".tar.gz"
+		}
+		// Upload file
+		url, err := upload(path, filename)
+		fmt.Println(url)
+		if err != nil {
+			return err
+		}
+		if ts.emitStdout {
+			log.Infof("Emitting stdout event")
```
walterchris, line 132: Idk if we need this logging, probably too verbose.

plugins/teststeps/uploadFile/uploadFile.go
```diff
@@ -0,0 +1,281 @@
...
+			if err != nil {
+				log.Fatalln("Error creating archive:", err)
+			}
+			fmt.Println("Tar archive created successfully")
+			path = newpath
+			filename = filename + ".tar.gz"
+		}
+		// Upload file
+		url, err := upload(path, filename)
+		fmt.Println(url)
```
walterchris, line 127: verbose

plugins/teststeps/uploadFile/uploadFile.go
```diff
@@ -0,0 +1,281 @@
...
+	}
+	return nil
+}
+
+// Upload the file that is specified in the JobDescritor
+func upload(path string, filename string) (string, error) {
+	// Read the file that should be uploaded
+	bodyBytes, err := ioutil.ReadFile(path)
+	if err != nil {
+		return "Could not read the file.", err
```
walterchris, line 242: I don't think that you need to return a string if an error occurred.

walterchris (changes requested):

plugins/teststeps/fileupload/awsFileUpload.go
```diff
@@ -0,0 +1,279 @@
...
+const (
+	EventURLtoReport = event.Name("URLtoReport")
+)
+
+// Events defines the events that a TestStep is allow to emit
+var Events = []event.Name{
+	EventURLtoReport,
+}
+
+// Cmd is used to run arbitrary commands as test steps.
```
tfg13, line 55: copy-paste

plugins/teststeps/fileupload/awsFileUpload.go
```diff
@@ -0,0 +1,279 @@
...
+	"github.com/facebookincubator/contest/pkg/xcontext"
+	"github.com/facebookincubator/contest/plugins/teststeps"
+	"github.com/prometheus/common/log"
+)
+
+// Define some constants for uploading things into a S3 bucket
+const (
+	S3_REGION = "" // your aws server region
+	S3_BUCKET = "" // your aws bucket
+	S3_PATH   = "" // your path where to upload the file
```
tfg13, line 39: Shouldn't this be parameters of the plugin? Do I need to change this file to use this plugin?
(Or maybe the plugin should read this from some file?)

plugins/teststeps/fileupload/awsFileUpload.go
tfg13, line 55: copy paste

plugins/teststeps/fileupload/awsFileUpload.go
```diff
@@ -0,0 +1,279 @@
...
+			}
+			defer tar_gz.Close()
+			// Create the archive and write the output to the "out" Writer
+			err = createTarArchive(path, tar_gz)
+			if err != nil {
+				log.Fatalln("Error creating archive:", err)
+			}
+			fmt.Println("Tar archive created successfully")
+			path = newpath
+			filename = filename + ".tar.gz"
```
tfg13, line 123: Where will it put the compressed file, next to the original uncompressed one?
I guess that is ok. I was just wondering, since later we have to read the entire file into memory for upload anyway, should we just write the archive into a byte buffer and then upload it, so basically without creating a `tar.gz` file that needs to be cleaned up? I think that would be nicer

tfg13 (changes requested):
Overall looking good, some questions

plugins/teststeps/awsfileupload/awsFileUpload.go
```diff
@@ -0,0 +1,300 @@
...
+		if ts.compress {
+			// Create buffer for the compressed data
+			var buf bytes.Buffer
+			// Create the archive and write the output to the "out" Writer
+			buf, err = createTarArchive(path, buf)
+			if err != nil {
+				return fmt.Errorf("error creating an archive: %v", err)
+			}
+			fmt.Println("Tar archive created successfully")
+			path = path + ".tar.gz"
```
walterchris, line 114: unused

plugins/teststeps/awsfileupload/awsFileUpload.go
```diff
@@ -0,0 +1,300 @@
...
+	// Write the file into the tar archive
+	_, err = io.Copy(tarwriter, file)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// Upload the file that is specified in the JobDescritor
+func (ts *FileUpload) upload(path string, filename string, gzbuffer []byte) (string, error) {
```
walterchris, line 266: Path is also unused as we're using the byte buffer now - right?

walterchris (changes requested):

plugins/teststeps/awsfileupload/awsFileUpload.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+	return nil
+}
+
+// Upload the file that is specified in the JobDescritor
+func (ts *FileUpload) upload(filename string, gzbuffer []byte) (string, error) {
+	// Create a single AWS session (we can re use this if we're uploading many files)
+	s, err := session.NewSession(&aws.Config{Region: aws.String(ts.s3region),
+		Credentials: credentials.NewSharedCredentials(
+			"", // your credential file path (default if empty)
```
walterchris, line 269: Do we have to fill in here something? Probably the options that you handed in? TBH I don't know what that is good for.

insomniacslk:
I haven't read the code and will let others review, but one thing I noticed is that the plugin name is too generic for what it does. Should this be called `s3fileuploader` or `s3upload` instead?

insomniacslk:
ah I just noticed Chris's comment and the fact that this was renamed `awsfileupload`. I would say that `aws` is still too generic, there are hundreds of AWS services, and I would just use `s3` in the name instead.

The PR title and commit messages still reference the old name, please change them as well, and squash the commits into one before merging

plugins/teststeps/awsfileupload/awsFileUpload.go
```diff
@@ -0,0 +1,299 @@
...
+
+// Events defines the events that a TestStep is allow to emit
+var Events = []event.Name{
+	EventURLtoReport,
+}
+
+// FileUpload is used to retrieve all the parameter, the plugin needs.
+type FileUpload struct {
+	path            *test.Param // Path to file that shall be uploaded
+	filename        *test.Param // Filename to file that shall be uploaded
```
insomniacslk, line 49: can this be joined to the `path` variable above? Then you can use `filepath.Base` and `filepath.Dir` to split it into its components in the code below

plugins/teststeps/awsfileupload/awsFileUpload.go
```diff
@@ -0,0 +1,299 @@
...
+	s3region        string      // AWS server region that shall be used
```
insomniacslk, line 50: is this field mandatory? If not can you add it to the docstring explaining what's the default otherwise?

plugins/teststeps/awsfileupload/awsFileUpload.go
```diff
@@ -0,0 +1,299 @@
...
+	s3bucket        string      // AWS Bucket name
+	s3path          string      // Path where in the bucket to upload the file
+	s3credfile      string      // Path to the AWS credential file to authenticate the session
+	s3credprofile   string      // Profile of the AWS credential file that shall be used
+	emitURLtoReport bool        // Bool that defines that URLtoReport Event shall be triggered
+	compress        bool        // Bool that defines if the data should be compressed before uploading it
```
insomniacslk, line 56: please document which type of compression is supported

plugins/teststeps/awsfileupload/awsFileUpload.go
```diff
@@ -0,0 +1,299 @@
...
+				return fmt.Errorf("could not upload the file: %v", err)
+			}
+		} else {
+			// Read the file that should be uploaded
+			bodyBytes, err := ioutil.ReadFile(path)
+			if err != nil {
+				return fmt.Errorf("could not read the file: %v", err)
+			}
+			// Upload file
+			url, err = ts.upload(filename, bodyBytes)
```
insomniacslk, line 127: nit: small duplication that could be removed

plugins/teststeps/awsfileupload/awsFileUpload.go
```diff
@@ -0,0 +1,299 @@
...
+			if err != nil {
+				return fmt.Errorf("could not upload the file: %v", err)
+			}
+		}
+		// Emit URLtoReport event to get the url into the report
+		if ts.emitURLtoReport {
```
insomniacslk, line 133: I think this should be done every time, without gating it behind a parameter. What would be a reason not to emit this event?

plugins/teststeps/awsfileupload/awsFileUpload.go
```diff
@@ -0,0 +1,299 @@
...
+	return teststeps.ForEachTarget(Name, ctx, ch, f)
+}
+
+// Retrieve all the parameters defines through the jobDesc
+func (ts *FileUpload) validateAndPopulate(params test.TestStepParameters) error {
+	// Retrieving parameter as json Raw.Message
+	// validate path and filename
+	ts.path = params.GetOne("path")
+	if ts.path.IsEmpty() {
+		return fmt.Errorf("invalid or missing 'path' parameter, must be exactly one string")
```
insomniacslk, line 149: a better error message, here and below, could be "missing or empty 'path' parameter". `IsEmpty` returns `true` if the parameter is not specified, or if it's an empty string, and "must be exactly one string" could be misleading

plugins/teststeps/awsfileupload/awsFileUpload.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+	ts.s3bucket = param.String()
+	// validate s3path
+	param = params.GetOne("s3path")
+	if param.IsEmpty() {
+		return fmt.Errorf("missing 's3path' parameter, must be exactly one string")
+	}
+	ts.s3path = param.String()
+	// retrieve s3credfile
+	ts.s3credfile = params.GetOne("s3credfile").String()
```
insomniacslk, line 176: If s3credfile and s3credprofile are empty, will the plugin fail or work?

plugins/teststeps/awsfileupload/awsFileUpload.go
```diff
@@ -0,0 +1,299 @@
...
+			return fmt.Errorf("invalid non-boolean `emit_urltoreport` parameter: %v", err)
+		}
+		ts.compress = v
+	}
+	return nil
+}
+
+// ValidateParameters validates the parameters associated to the TestStep
+func (ts *FileUpload) ValidateParameters(_ xcontext.Context, params test.TestStepParameters) error {
+	return nil
```
insomniacslk, line 204: this must call `validateAndPopulate`

plugins/teststeps/awsfileupload/awsFileUpload.go
```diff
@@ -0,0 +1,299 @@
...
+}
+
+// addFileToArchive takes the data and writes it into the tar archive
+func addFileToArchive(tarwriter *tar.Writer, filename string) error {
+	// Open the file which shall be written into the tar archive
+	file, err := os.Open(filename)
+	if err != nil {
+		return err
+	}
+	defer file.Close()
```
insomniacslk, line 239: `file.Close()` should be checked, it can fail in significant ways (I can see Tobias chuckle at this comment :D ). For example:
```
defer func() {
    if err := file.Close(); err != nil {
         ctx.Logger().Warnf("failed to close file '%s': %w", filename, err)
    }
}
```

plugins/teststeps/awsfileupload/awsFileUpload.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+	return buf, nil
+}
+
+// addFileToArchive takes the data and writes it into the tar archive
+func addFileToArchive(tarwriter *tar.Writer, filename string) error {
+	// Open the file which shall be written into the tar archive
+	file, err := os.Open(filename)
+	if err != nil {
+		return err
```
insomniacslk, line 237: wrap this error message with something that will help the user to understand why this function has failed. E.g.
```
return fmt.Errorf("os.Open failed: %w", err)
```

or, if the caller doesn't print the file name (but it should):
```
return fmt.Errorf("failed to open '%s': %w", filename, err)
```

plugins/teststeps/awsfileupload/awsFileUpload.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+	defer file.Close()
+	// Retrieve the file stats
+	info, err := file.Stat()
+	if err != nil {
+		return err
```
insomniacslk, line 243: same here and below about wrapping the error message with important information

plugins/teststeps/awsfileupload/awsFileUpload.go
```diff
@@ -0,0 +1,299 @@
...
+// Upload the file that is specified in the JobDescritor
+func (ts *FileUpload) upload(filename string, gzbuffer []byte) (string, error) {
+	// Create a single AWS session (we can re use this if we're uploading many files)
+	s, err := session.NewSession(&aws.Config{Region: aws.String(ts.s3region),
+		Credentials: credentials.NewSharedCredentials(
+			"", // your credential file path (default if empty)
+			"", // profile name (default if empty)
+		)})
+	if err != nil {
+		return "Could not open a new session.", err
```
insomniacslk, line 273: fix to 
```
return fmt.Errorf("could not open a new session: %w", err)
```

note that error strings in Go should never start with a capital letter

plugins/teststeps/awsfileupload/awsFileUpload.go
```diff
@@ -0,0 +1,299 @@
...
+	// Write the file into the tar archive
+	_, err = io.Copy(tarwriter, file)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// Upload the file that is specified in the JobDescritor
+func (ts *FileUpload) upload(filename string, gzbuffer []byte) (string, error) {
```
insomniacslk, line 265: nit: `gzbuffer` is misnamed. This function has no knowledge of whether this is compressed or not, just call it `data`

plugins/teststeps/awsfileupload/awsFileUpload.go
```diff
@@ -0,0 +1,299 @@
...
+		Body:                 bytes.NewReader(gzbuffer),
+		ContentLength:        aws.Int64(int64(len(gzbuffer))),
+		ContentType:          aws.String(http.DetectContentType(gzbuffer)),
+		ContentDisposition:   aws.String("attachment"),
+		ServerSideEncryption: aws.String("AES256"),
+	})
+	if err != nil {
+		return "", err
+	} else {
+		fmt.Printf("Pushed the file to S3 Bucket! \n")
```
insomniacslk, line 294: no `fmt`, use `ctx.Logger().Infof()`

plugins/teststeps/awsfileupload/awsFileUpload.go
```diff
@@ -0,0 +1,299 @@
...
+		Key:                  aws.String(fileName),
+		ACL:                  aws.String("public-read"),
+		Body:                 bytes.NewReader(gzbuffer),
+		ContentLength:        aws.Int64(int64(len(gzbuffer))),
+		ContentType:          aws.String(http.DetectContentType(gzbuffer)),
+		ContentDisposition:   aws.String("attachment"),
+		ServerSideEncryption: aws.String("AES256"),
+	})
+	if err != nil {
+		return "", err
```
insomniacslk, line 292: wrap the error message

plugins/teststeps/awsfileupload/awsFileUpload.go
```diff
@@ -0,0 +1,299 @@
...
+	} else {
+		fmt.Printf("Pushed the file to S3 Bucket! \n")
+	}
+	// Create download link for public ACL
+	url := "https://" + ts.s3bucket + ".s3.amazonaws.com/" + fileName
```
insomniacslk, line 297: this might change in the future - doesn't the API call to S3 return the URL, or its components? You can then join them with `url.URL`

insomniacslk (changes requested):

plugins/teststeps/awsfileupload/awsFileUpload.go
tfg13: Aaaaaaah!

plugins/teststeps/awsfileupload/awsFileUpload.go
tfg13: `ctx.Infof()` also works and is shorter :)

plugins/teststeps/awsfileupload/awsFileUpload.go
llogen: Thank you for the hint. I can be used if you manage multiple profiles in your aws credential file. Also your can adapt the path to your aws credential file. If not set it uses defaults.

plugins/teststeps/awsfileupload/awsFileUpload.go
llogen: Unless there is no config file that defines the region it is mandatory. There is no default otherwise.

plugins/teststeps/awsfileupload/awsFileUpload.go
llogen: If they are empty, the session will be open with the default aws credential file path and the default credential profile. But this default also have to be created in order to authenticate to an aws session. So the answer is if this credential file misses, the plugin will always fail. If it is stored as default, the plugin will work, even if the strings are empty.

plugins/teststeps/awsfileupload/awsFileUpload.go
llogen: No the API just return an ETag, the ServerSideEncryption and the VersionId. This is why I put it together like that.

plugins/teststeps/awsfileupload/awsFileUpload.go
llogen: I think it is fine with localpath and filename. The validation is very handy like this.

rojer9-fb:
please rename the file to s3fileupload.go to align with the local naming convention

plugins/teststeps/s3fileupload/s3FileUpload.go
```diff
@@ -0,0 +1,289 @@
...
+)
+
+// Events defines the events that a TestStep is allow to emit
+var Events = []event.Name{
+	EventURL,
+}
+
+// FileUpload is used to retrieve all the parameter, the plugin needs.
+type FileUpload struct {
+	localpath     *test.Param // Path to file that shall be uploaded
```
rojer9-fb, line 48: field names (unlike files) should use camelCase, hence `localPath`, `fileName`, etc.

plugins/teststeps/s3fileupload/s3FileUpload.go
```diff
@@ -0,0 +1,289 @@
...
+			ts.s3credfile,    // your credential file path (default if empty)
+			ts.s3credprofile, // profile name (default if empty)
+		)})
+	if err != nil {
+		return "", fmt.Errorf("could not open a new session: %w", err)
+	}
+	currentTime := time.Now()
+	// Config settings: this is where you choose the bucket, filename, content-type etc.
+	// of the file you're uploading.
+	fileName := fmt.Sprintf("%s/%s_%s", ts.s3path, currentTime.Format("20060102_150405"), filename)
```
rojer9-fb, line 268: use filepath.Join to join file paths

plugins/teststeps/s3fileupload/s3fileupload.go
```diff
@@ -0,0 +1,291 @@
...
+		Body:                 bytes.NewReader(data),
+		ContentLength:        aws.Int64(int64(len(data))),
+		ContentType:          aws.String(http.DetectContentType(data)),
+		ContentDisposition:   aws.String("attachment"),
+		ServerSideEncryption: aws.String("AES256"),
+	})
+	if err != nil {
+		return "", fmt.Errorf("could not upload the file: %w", err)
+	} else {
+		ctx.Infof("Pushed the file to S3 Bucket! \n")
```
rojer9-fb, line 286: remove `\n` at the end, it'll just insert an empty line

rojer9-fb:
i think we should import and register the plugin into the default binary, so it's at least build-tested in the main repo

PR #299, opened 2021-08-18
make sure jobID, runID are available in Reporters, TestSteps, TMs

tfg13:
... and switch Reporters to use xcontext to allow this

This means TestSteps, Reporters and TargetManagers can now find out about their own `JobID` and `RunID` with a simple

`types.JobIDFromContext(ctx)`

Also adding integration tests that make sure these values are actually available so this won't regress.

### DIFF
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -5,7 +5,11 @@
 
 package types
 
-import "strconv"
+import (
+	"strconv"
+
+	"github.com/facebookincubator/contest/pkg/xcontext"
+)
 
 // JobID represents a unique job identifier
 type JobID uint64
@@ -19,4 +23,28 @@ func (v JobID) String() string {
 
 func (v RunID) String() string {
 	return strconv.FormatUint(uint64(v), 10)
+}
+
+type key string
+
+const (
+	KeyJobID = key("job_id")
+	KeyRunID = key("run_id")
+)
+
+// JobIDFromContext is a helper to get the JobID, this is useful
+// for plugins which need to know which job they are running.
+// Not all context object everywhere have this set, but this is
+// guaranteed to work in TargetManagers, TestSteps and Reporters
+func JobIDFromContext(ctx xcontext.Context) (JobID, bool) {
+	v, ok := ctx.Value(KeyJobID).(JobID)
+	return v, ok
+}
+
+// RunIDFromContext is a helper to get the RunID.
+// Not all context object everywhere have this set, but this is
+// guaranteed to work in TargetManagers, TestSteps and RunReporters
+func RunIDFromContext(ctx xcontext.Context) (RunID, bool) {
+	v, ok := ctx.Value(KeyRunID).(RunID)
+	return v, ok
 }
--- a/tests/plugins/reporters/readmeta/readmeta.go
+++ b/tests/plugins/reporters/readmeta/readmeta.go
@@ -1,0 +1,72 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+//
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+package readmeta_test
+
+import (
+	"github.com/facebookincubator/contest/pkg/event/testevent"
+	"github.com/facebookincubator/contest/pkg/job"
+	"github.com/facebookincubator/contest/pkg/types"
+	"github.com/facebookincubator/contest/pkg/xcontext"
+)
+
+// Name defines the name of the reporter used within the plugin registry
+var Name = "readmeta"
+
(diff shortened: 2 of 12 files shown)


tests/plugins/reporters/readmeta/readmeta.go
```diff
@@ -0,0 +1,72 @@
...
+	return true, "I did nothing", nil
+}
+
+// FinalReport calculates the final report to be associated to a job.
+func (n *Readmeta) FinalReport(ctx xcontext.Context, parameters interface{}, runStatuses []job.RunStatus, ev testevent.Fetcher) (bool, interface{}, error) {
+	// this one only has jobID, there is no specific runID in the final reporter
+	jobID, ok1 := types.JobIDFromContext(ctx)
+	// note this must use panic to abort the test run, as this is a test for the job runner which ignore the actual outcome
+	if jobID == 0 || !ok1 {
+		panic("Unable to extract jobID from context")
```
rojer9-fb, line 59: shouldn't be panicing when it's possible to return an error

tests/plugins/targetmanagers/readmeta/readmeta.go
```diff
@@ -0,0 +1,75 @@
...
+// ValidateReleaseParameters valides parameters that will be passed to Release.
+func (r Readmeta) ValidateReleaseParameters(params []byte) (interface{}, error) {
+	return nil, nil
+}
+
+// Acquire implements contest.TargetManager.Acquire
+func (t *Readmeta) Acquire(ctx xcontext.Context, jobID types.JobID, jobTargetManagerAcquireTimeout time.Duration, parameters interface{}, tl target.Locker) ([]*target.Target, error) {
+	// test metadata exists
+	jobID2, ok1 := types.JobIDFromContext(ctx)
+	// note this must use panic to abort the test run, as this is a test for the job runner which ignore the actual outcome
```
rojer9-fb, line 39: you can emit an event with job id and run id and verify by fetching them afterwards (there's a helper for that)

pkg/types/types.go
```diff
@@ -20,3 +24,20 @@ func (v JobID) String() string {
 func (v RunID) String() string {
 	return strconv.FormatUint(uint64(v), 10)
 }
+
+// JobIDFromContext is a helper to get the JobID, this is useful
+// for plugins which need to know which job they are running.
+// Not all context object everywhere have this set, but this is
+// guaranteed to work in TargetManagers, TestSteps and Reporters
+func JobIDFromContext(ctx xcontext.Context) (JobID, bool) {
+	v, ok := ctx.Value("job_id").(JobID)
```
xaionaro, line 33: You need to define a special type for the key and use it. This way you will guarantee no accidental clashing:
```
type jobIDKey struct{}

ctx = ctx.WithValue(jobIDKey, jobID)
```

tests/plugins/reporters/readmeta/readmeta.go
tfg13: This will make the test code more complex - do you think that is worth it? I would need to get the status/report and see if the job was successful, which this test is currently not doing.

tests/plugins/targetmanagers/readmeta/readmeta.go
tfg13: Are you sure I can emit events from the target manager and/or reporter?
(test step works of course)

pkg/types/types.go
tfg13: TIL. Fixed

pkg/types/types.go
```diff
@@ -20,3 +24,27 @@ func (v JobID) String() string {
...
+)
+
+// JobIDFromContext is a helper to get the JobID, this is useful
+// for plugins which need to know which job they are running.
+// Not all context object everywhere have this set, but this is
+// guaranteed to work in TargetManagers, TestSteps and Reporters
+func JobIDFromContext(ctx xcontext.Context) (JobID, bool) {
+	v, ok := ctx.Value(KeyJobID).(JobID)
+	return v, ok
+}
```
xaionaro, line 42: I understand this sounds very controversial, and personally I do not really have an opinion. But...

I somewhat do not like the idea of checking `ok` **each time** we extract `JobID`. So, may be panicking is fine in this case?

Do not take this comment seriously, it is just a crazy thought.

tests/plugins/targetmanagers/readmeta/readmeta.go
tfg13: Done in the test step, I think this is the only place that can easily do it

pkg/types/types.go
tfg13: Yeah I also though about this. I ended up with this:
```
// guaranteed to work here
jobID, _ := types.JobIDFromContext(ctx)
```

PR #300, opened 2021-08-20
pluginregistry: Better error for unknown teststeps

tfg13:
When creating a job, we load the allowed Events before the test step itself, which means unknown test step result in this confusing error message:
```
"Err": "could not create step bundles: could not create test steps bundle: TestStep foobar does not have any event associated"
```

Change this around to load the test step first, now the error message is:
```
"Err": "could not create step bundles: could not create test steps bundle: NewStepBundle for test step 'foobar' with index 0 failed: could not get the desired TestStep (foobar): TestStep foobar is not registered"
```

### DIFF
--- a/pkg/jobmanager/bundles.go
+++ b/pkg/jobmanager/bundles.go
@@ -68,16 +68,9 @@ func newBundlesFromSteps(ctx xcontext.Context, descriptors []*test.TestStepDescr
 		if err := limits.NewValidator().ValidateTestStepLabel(descriptor.Label); err != nil {
 			return nil, err
 		}
-		tse, err := registry.NewTestStepEvents(descriptor.Name)
-		if err != nil {
-			return nil, err
-		}
-		tsb, err := registry.NewTestStepBundle(ctx, *descriptor, tse)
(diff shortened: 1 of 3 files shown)


pkg/jobmanager/bundles.go
```diff
@@ -68,17 +68,10 @@ func newBundlesFromSteps(ctx xcontext.Context, descriptors []*test.TestStepDescr
...
-		tse, err := registry.NewTestStepEvents(descriptor.Name)
-		if err != nil {
-			return nil, err
-		}
-		tsb, err := registry.NewTestStepBundle(ctx, *descriptor, tse)
+		tsb, err := registry.NewTestStepBundle(ctx, *descriptor)
 		if err != nil {
 			return nil, fmt.Errorf("NewStepBundle for test step '%s' with index %d failed: %w", descriptor.Name, idx, err)
 		}
-		if err != nil {
```
tfg13, line 79: this looks like a merge artifact, it is unreachable

insomniacslk (approved):
That error message has annoyed me for such a long time :D thanks for fixing it!

